PR #5, opened 2018-10-04
Fix sign and nuclear charge in Coulomb correction

straitm:
The Coulomb correction for the outgoing nucleus needs to reverse
sign for mu+.  Also the relevant nuclear charge is that of the
residual nucleus, not the target nucleus.  Both fixed.

### DIFF
--- a/src/Physics/Common/PrimaryLeptonGenerator.cxx
+++ b/src/Physics/Common/PrimaryLeptonGenerator.cxx
@@ -173,7 +173,11 @@ void PrimaryLeptonGenerator::AddToEventRecord(
         << "Correcting f/s lepton energy for Coulomb effects";
 
     double m = interaction->FSPrimLepton()->Mass();
-    double Z  = nucltgt->Z();
+
+    double z = pdgc > 0? -1: +1; // charge of the lepton
+
+    // Get Z of residual nucleus, assuming that all nucleons are still present
+    double Z  = nucltgt->Z() - z;
     double A  = nucltgt->A();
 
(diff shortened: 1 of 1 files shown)


sjgardiner:
Thanks, Matthew, for your pull request, and sorry for the long delay in providing feedback. I agree with your changes to the calculation of the Coulomb-corrected lepton energy. However, I think that there is an error in the overall approach used by `src/Physics/Common/PrimaryLeptonGenerator.cxx`.

My understanding is that the corrected energy (as described in, e.g., [Engel's paper](https://arxiv.org/pdf/nucl-th/9711045.pdf) on the effective momentum approximation and [Bodek's paper](https://arxiv.org/pdf/1801.07975.pdf) on removal energies) applies only to the charged lepton when it is _within the nucleus_. Therefore, the corrected energy should be used when calculating the cross section (see Engel's work) and when reconstructing CCQE events using only the lepton kinematics (see Bodek's). However, I believe that the energy of the charged lepton far from the nucleus (where the Coulomb energy may be taken to be zero) is the correct value to include in the event record.

If my interpretation is accurate, then this sort of correction likely belongs in a cross section model (see, e.g., the existing Coulomb correction available in `src/Physics/QuasiElastic/XSection/NievesQELCCPXsec.cxx`) instead of the event generator. @dytman and @candreop, what do you think?

candreop:
I believe your interpretation is correct Steven.

mroda88:
This pull request was stuck as there was no consensus. Since the changes are minimal we close it and we open an issue to remember this discussion.

PR #7, opened 2018-10-10
Rewrite the genie-write-gversion script to use git instead of svn

sjgardiner:
I've rewritten the genie-write-gversion script to be compatible with git. The biggest change is that `__GENIE_SVN_REVISION__` is no longer defined. Instead, `__GENIE_GIT_REVISION__` is a preprocessor macro which evaluates to a string containing the git hash of the current commit. A full `git diff HEAD` (which shows all uncommitted changes that existed at compilation time) is also given as an extended comment in the `GVersion.h` header file.

### DIFF
--- a/src/scripts/setup/genie-write-gversion
+++ b/src/scripts/setup/genie-write-gversion
@@ -4,6 +4,7 @@
 #
 # C.Andreopoulos <costas.andreopoulos \at stfc.ac.uk>, Rutherford Lab.
 #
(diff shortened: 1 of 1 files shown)


sjgardiner:
If we really want the current commit count (SVN style) instead of the commit hash, we could use `git rev-list --count HEAD` in the script instead of `git rev-parse --verify HEAD`. I'd advise against this, though, since it is not sensitive to which branch is being used.

Issue #3, opened 2018-10-03
`Cache::Save()` passes wrong arguments to `TList::Write()`


PetrilloAtWork:
GNU GCC 7.3.0 compiler reports:
~~~~
Cache.cxx: In member function ‘void genie::Cache::Save()’:
Cache.cxx:186:57: warning: enum constant in boolean context [-Wint-in-bool-context]
   keys->Write("key_list", TObject::kSingleKey&&TObject::kOverwrite);
                                                         ^~~~~~~~~~
~~~~
https://github.com/GENIE-MC/Generator/blob/cf98cb6ebd44fe5ae22b05d648a22bac7fcb9393/src/Framework/Utils/Cache.cxx#L186

The second argument should be a bit mask; I suppose the intention was to have `TObject::kSingleKey | TObject::kOverwrite` (use a single key and overwrite it as needed).
The `&&` operator makes teh expression a boolean (`1 && 2`), that is equivalent to `1`, that is just `TObject::kSingleKey`, and no overwrite.

mroda88:
Fixed, thank you for the catch!

PR #10, opened 2018-10-29
Refactor Nieves MEC model to use the new hadron tensor interface

sjgardiner:
As part of an ongoing effort to add new models based on spectral functions (O. Benhar, N. Rocco, et al.), superscaling (SuSAv2 by G. Megias and collaborators), etc. to GENIE, I've developed a generalized interface for describing hadron tensors. I presented a few [slides](https://indico.fnal.gov/event/18545/session/1/material/0/3.pdf) pitching the idea at the last [Fermilab Neutrino Joint Theory-Experiment Working Group](http://npc.fnal.gov/neutrino-joint-theory-experiment-working-group/) meeting.

Before adding in new physics models, I thought it would make sense to rewrite the Nieves MEC cross section model (which uses precomputed hadron tensor tables) as a proof of concept.

Some key changes from the original implementation include:

1. The hadron tensor tables are now self-describing, with a header listing the target nuclide, grid spacing in q0 and q3, etc. (Written documentation is forthcoming)

2. The new `HadronTensorPool` singleton class is an XML-configurable manager for all hadron tensor objects. It replaces the old `MECHadronTensor` class with a more flexible design.

3. The tensors themselves are described by a new abstract base class, `HadronTensorI`. The `ValenciaHadronTensorI` derived class is also abstract, but it conforms to the Valencia group's conventions for the structure functions, etc. The concrete derived class `TabulatedValenciaHadronTensor` uses precomputed tables to evaluate the elements of the hadron tensor. In future versions of the interface, additional concrete derived classes may be added to calculate them on demand.


sjgardiner:
As a first check, here's a comparison of numu CCMEC splines for 12C before and after the refactoring: screen shot 2018-10-29 at 12 35 58 pm

candreop:
Thanks Stephen! This is a very useful development. Also, the implementation looks to be at a very high standard and is very clean. I am very keen to integrate this in GENIE asap. But we never discussed the development in any detail (requirements, design, validation plan) within GENIE, and I want to avoid outsourcing important GENIE development discussions to the Fermilab Neutrino Joint Theory-Experiment Working Group. This is not a GENIE group, and I (and many others) do not have the time to attend the mtgs of all such local, interest groups at Fermilab, CERN or elsewhere. So we need to set this up as a GENIE incubator project, and give the opportunity to all core GENIE developers (as opposed to Fermilab-based users, bystanders and other neutrino interaction afficionados) to have a detailed look. This can be expedited, given the maturity of the development, but I want to see some thought and discussion about the requirements and, in particular, some thinking on a) modelling uncertainties within that framework, and b) how we capture information about what physics is included in the hadron tensor so that we can communicate that information with other segments of the GENIE code (models or event generation mechanics).

candreop:
Hi Stephen. Please can you think of your development as an incubator project (or a series of projects) and add a page on the GENIE DocDB with the scope and goals? If the scope of the devel is to develop a generator/theory interface then we probably need to discuss this more and agree on how the interface should look like (before we advertise it as such to the theory community). Ultimately, I think it is undesirable to have a theory interface which relies on hadron tensors alone (although hadron tensors might well be part of the required mix). It reduces the generator to a collection of look-up tables and makes it nearly impossible to tune and assess systematics. But if the scope is to refactorize and improve some of the current code (the part of it that uses hadron tensors anyhow) then I see this as an easier win and a much less debatable development. We could clear this in the next few days so we could focus on the bigger picture.

sjgardiner:
Hi Costas, thanks very much for the feedback! I've talked a bit with Gabe recently, and I think I now have a better understanding of the incubator project development model. I agree that it makes sense to limit the initial scope of this project pending further discussion within GENIE. Perhaps the right way to organize future efforts in this direction is to have a series of 3 incubator projects: (1) refactoring the Nieves MEC model to use a cleaner interface for managing the tables of hadron tensors (essentially what you see in the current pull request), (2) adding a similar implementation for the SuSAv2-MEC model, with the intent to make as many pieces tunable/reweightable as possible (this incubator project already exists), and (3) adding a new generator/theory interface to GENIE to facilitate the incorporation of new models. As you point out, #1 should be relatively uncontroversial, although some additional validation should be done. I'd like to discuss my thoughts on #3 with you and other GENIE collaborators sometime soon. Ultimately I see the hadron tensor tables as a crutch to use for an initial implementation. Fully tunable calculations done within GENIE are the ultimate goal.

If that series of proposals sounds reasonable to you, I'm happy to write up descriptions of these incubator projects for the GENIE DocDB. We could then discuss them as a collaboration whenever would be most convenient, e.g., during the next Monday phone call.

candreop:
Hi Steve. I agree unreservedly with everything you said above.

PR #13, opened 2018-11-27
Replace "DipoleELFF" with "Dipole" in configuration for the NievesQELCCPXSec model

sjgardiner:
As discussed with @dytman and @nusense today, the Nieves CCQE model uses the name "DipoleELFF" for its dipole form factor configuration instead of "Dipole," which is used by other models and is assumed to be present by some existing reweight calculators. This pull request corrects the discrepancy.

### DIFF
--- a/config/G18_10a/ModelConfiguration.xml
+++ b/config/G18_10a/ModelConfiguration.xml
@@ -66,7 +66,7 @@ STFC, Rutherford Appleton Laboratory
 <!-- New Nieves QE model   -->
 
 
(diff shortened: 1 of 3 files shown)


mroda88:
Steve, can you replace the string in every XML file? You only replaced in files where the string is used, but I believe it is in every sub directory of config,  although commented out. Can you replace all of them?

sjgardiner:
I can't see any more instances of "DipoleELFF" anywhere (`grep -r` doesn't show anything). If there are more on the latest master branch that I've somehow missed, please feel free to change them.

PR #14, opened 2018-11-30
Phase space fix for differential xsec stored by QELEventGenerator

sjgardiner:
Currently, the differential cross section stored in the event record by QELEventGenerator is claimed to be in the kPSQ2fE phase space, but checks performed while reweighting MaCCQE demonstrate that this is not the case. This leads to unphysically large systematic errors in the GReWeightNuXSecCCQE calculator. The problem is resolved by storing the correct xsec value in this phase space.

### DIFF
--- a/src/Physics/QuasiElastic/EventGen/QELEventGenerator.cxx
+++ b/src/Physics/QuasiElastic/EventGen/QELEventGenerator.cxx
@@ -318,6 +318,15 @@ void QELEventGenerator::ProcessEventRecord(GHepRecord * evrec) const
 
         // If the generated kinematics are accepted, finish-up module's job
         if(accept) {
+
+            // Stored differential cross section uses kPSQ2fE phase space,
+            // but xsec as calculated above does not match. We therefore
+            // recalculate the differential cross section for kPSQ2fE here.
+            // -- S. Gardiner
+            // set the cross section for the selected kinematics
+            double stored_xsec = fXSecModel->XSec(interaction, kPSQ2fE);
+            evrec->SetDiffXSec(stored_xsec,kPSQ2fE);
+
             double gQ2 = interaction->KinePtr()->Q2(false);
             LOG("QELEvent", pINFO) << "*Selected* Q^2 = " << gQ2 << " GeV^2";
 
@@ -361,9 +370,6 @@ void QELEventGenerator::ProcessEventRecord(GHepRecord * evrec) const
             interaction->KinePtr()->Setx (gx,  true);
             interaction->KinePtr()->Sety (gy,  true);
             interaction->KinePtr()->ClearRunningValues();
-
-            // set the cross section for the selected kinematics
-            evrec->SetDiffXSec(xsec,kPSQ2fE);
 
             TLorentzVector lepton(interaction->KinePtr()->FSLeptonP4());
             TLorentzVector outNucleon(interaction->KinePtr()->HadSystP4());


sjgardiner:
You're right. The procedure I'm suggesting here is a band-aid instead of an optimal solution. As you mention, the right approach would be to report the correct phase space and adjust the reweighting code (which assumes `kPSQ2fE`) appropriately. There's an issue with doing so, however. In the current version of QELEventGenerator, after computing the differential cross section in the `kPSTnctnBnctl` phase space,
```
xsec = fXSecModel->XSec(interaction, kPSTnctnBnctl);
```
the ComputeXSec() method of QELEventGenerator then multiplies by some additional factors before doing the accept/reject test:
```
xsec *= 4*kPi*leptonCOM.P()*leptonCOM.P();
double jac = this->COMJacobian(lepton, leptonCOM, outNucleon, beta);
xsec *= jac;
```
The new value of `xsec` is then used unaltered both for kinematic sampling and when storing the differential cross section in the event record:
```
evrec->SetDiffXSec(xsec, kPSQ2fE);
```
Changing the reported phase space in the call to `SetDiffXSec()` is clearly the best way forward, but it's not obvious to me what the right enum value is, or whether we even have a value of `KinePhaseSpace_t` that corresponds to what's being used here. @candreop, are you familiar enough with this code to describe the intent behind applying the extra factors?

candreop:
No, this code is one of the small batch of codes that were introduced into GENIE when no one was looking closely and they are substandard. "It works" is not an acceptable argument and no such code will be allowed in again. If we can not easily see what the code does then, arguably, it does *not* work because we find ourselves in situation like this and can do our job. So, these factors is just a Jacobian. But for which kinematical tranformation? I am not going to sit and calculate the Jacobian for all relevant combinations of variables till I hit the transformation that gives me the same Jacobian as in the code. All it would have taken is a few lines of comments. I would assert this: Unless we actually store the actual x-section used for kinematical selection (in whatever space it was) we may not, actually, reweight the event properly! We need to make sure that all kinematical quantities that were selected are taken into account when we calculate weights. Doing a variable transfmation should be all right, but ignorning some variable is certainly not. I am bit worried because the code reports a single-differental x-section, but we select more than one independent variables. And here is the cherry on top of the cake: Forget v3 (that has not been used much just yet). If there is a mistake, it is likely that the same mistake existed in v2.12,  or even earlier versions... Did we just saw it now because of the extra checks installed to catch tune misconfiguration problems?

dytman:
I have worked closely with Steven on this situation.  It started when ReWeight complained that stored CCQE cross section is different than current, turned out to be 50-80% different.  The effect is that the MaCCQE systematic is significantly larger than what it should be. Steven tracked it down to phase space - what is used in stored event is different than real event.  Long term, we need to decide what is right.  Now, we need to find improved solution which is what we've done.  Steven assumes event generator code is right and makes ReWeight use that value.  I have no problem approving this request.

dytman:
comment to Costas.  Yes, it looks like this was a problem in the past.  Yes, we now see it because we were doing detailed checks.

candreop:
Well, no, Steve @dytman. Unless we are _convinced_ that what we do is right we can not approve a fix and tag a new ReWeight version to be used by major experiments for physics results! There is no "improved" solution. Reweighting in this case can be exact! So, there are only 2 options: What we do can be shown to be either correct or wrong. "Wrong by a smaller amount" is not going to satisfy us or the experiments since this is a systematic on a dominant process.

candreop:
This explains the 2.12 problem that was reported to me yesterday by Tarak Thakore during my visit to IFIC. Certainly, he was not the only person that was caught by this. If people still rely on 2.12.* for physics results, we may have to issue a 2.12 revision (in addition to fixing this in v3)

dytman:
I appreciate your point, Costas.  What we have now is inconsistency between what is generated and what is used in reweight.  As a result, the systematic associated with MaCCQE is clearly significantly wrong (overestimated).  We can make them consistent, longer term will allow us to decide what is right, perhaps neither of them. :)

dytman:
I should be more careful in my words.  It's a problem with both phase space and Jacobian.

candreop:
It is not a matter of consistency. The event generation code and event reweighting code could well use different kinematical variables. There is no unique approriate set and working with one or another set should be OK, as long as the Jacobian is calculated correctly and taken into account. The problem enters if one selects 2 independent variables but then calculates a weight ignoring one of them. Clearly, there is a correct thing to do here - this is one of the cases where reweighting can be exact. So we need to be convinced.

candreop:
You are confused here @dytman. You confuse a) kinematic variables with b) physics parameters. Yes, in this case we are discussing, we do reweight a single physics parameter within the same physics model. But, no, the weight is not necessarily constructed using the same differential cross-section as in event generation. I can select W and Q2 using d2sigma/dWdQ2 during event generation, but then I can reweight using d2sigma/dxdy. As long as I have all information and do the right transformation, there is nothing wrong with this. It is equivalent. The problem here is that we reweight based on dsigma/dQ2. And I ask again: Is it appropriate to reweight using a cross-section that is differential over a single variable? The event generation code selects more than 1 independent variables!

candreop:
Perhaps we should take the discussion to Slack. This thing here is visible to all users.

PR #21, opened 2019-02-23
Fix memory leaks.

igor144:
I tested the fix with Debian 7.0 ROOT5 and Debian 7.0 ROOT6 on my local PC, where it works properly.
But there was a crash on slf6 ROOT5 e9:prof and slf6 ROOT5 e9:debug
when **gevgen** was run on the test system: http://dbweb5.fnal.gov:8080/GenieCI/app/storage/docs/2018/12/11/stderr__Xg6MR2R.log.
The reason is unclear for me.

### DIFF
--- a/src/Physics/HadronTransport/INukeHadroData2018.cxx
+++ b/src/Physics/HadronTransport/INukeHadroData2018.cxx
@@ -75,6 +75,7 @@ INukeHadroData2018::INukeHadroData2018()
 //____________________________________________________________________________
 INukeHadroData2018::~INukeHadroData2018()
 {
+  
   // pi+n/p hA x-section splines
   delete fXSecPipn_Tot;
   delete fXSecPipn_CEx;
@@ -113,6 +114,7 @@ INukeHadroData2018::~INukeHadroData2018()
   // K+N x-section splines
   delete fXSecKpn_Elas;
   delete fXSecKpp_Elas;
+  delete fXSecKpn_CEx;
   delete fXSecKpN_Abs;
   delete fXSecKpN_Tot;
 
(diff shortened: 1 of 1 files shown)


igor144:
The important thing that is there is no such problem on slf6 ROOT6 e9:prof and slf6 ROOT6 e9:debug

nusense:
On Feb 23, 2019, at 5:39 PM, Igor Kakorin <notifications@github.com<mailto:notifications@github.com>> wrote:


The important thing that is there is no such problem on slf6 ROOT6 e9:prof and slf6 ROOT6 e9:debug

Only because you got lucky ... for that particular build, load, and run of GENIE ... not because it it correct code.  This is a common, but insidious, kind of bug with C++.

For instance (I haven't checked every case), but take  delete TPipA_Tot;

Search the code for where "TPipA_Tot" is initialized.   A simple grep shows:

$ grep -n TPipA_Tot INukeHadroData2018.*
INukeHadroData2018.cxx:817:    TPipA_Tot = new TGraph2D(pipATot_npoints);
...


But if you look more carefully with a context sensitive editor you find that


  if (saveTGraphsToFile) {
    string filename = "TGraphs.root";
    LOG("INukeHadroData2018", pNOTICE) << "Saving INTRANUKE hadron x-section data to ROOT file: " << filename;
    TGraphs_file.Open(filename.c_str(), "RECREATE");
  }

  /*   //<===== not the start of a commented out section
  // kIHNFtTot,   pip + A                                            PipA_Tot
  {
    const int pipATot_nfiles = 22;
    const int pipATot_nuclei[pipATot_nfiles] = {1, 2, 3, 4, 6, 7, 9, 12, 16, 27, 28, 32, 40, 48, 56, 58, 63, 93, 120, 165, 181, 209\
};
    const int pipATot_npoints = 203;

    TPipA_Tot = new TGraph2D(pipATot_npoints);    //<=== here's that line


what appears to be the initialization is commented out.   So when it gets around to deleting it ... that pointer is not necessarily zero (which is fine to "delete"), but possibly random garbage.   Which cases the crash happens depending on LOTS of external circumstances.   That it ran on one platform doesn't mean it will run on someone else's, even if they have the same setup on a similar platform.

Ideally all the member data pointers that you're deleting in the destructor should get initialized to zero in the constructor.   This prevents this sort of randomness.


-robert

Robert Hatcher
Computational Physics Developer

Scientific Computing Division
Fermi National Accelerator Laboratory
PO Box 500, MS 234, Batavia, IL 60510, USA
rhatcher@fnal.gov<mailto:rhatcher@fnal.gov>
Office: 630-840-3102<tel:630-840-3102>
Cell:    630-234-0091<tel:630-234-0091>

igor144:
Yes, the initialization of the pointer TPipA_Tot is commented, so it points to garbage.
But there are pointers which are initialized, but not deleted:
  fhN2dXSecKpN_CEx, TfracPipA_Abs, TfracPipA_CEx, TfracPipA_Inelas, 
  TfracPipA_PiPro, fFracKA_Tot, fFracKA_Elas, fFracKA_Inel, fFracKA_Abs.

So I agree with your suggestion to initialize all declared pointer to zero in the constructor and delete them in the destructor. It helps to avoid problems as with memory leaks as with segfaults when some part of the code will be commented out or deleted.

mroda88:
We need a new pull request with the proper bug fixes

PR #22, opened 2019-02-23
update resonance parameters in genie-pdg-table

igor144:
for consistency with BaryonResUtils.cxx according to last PDG.


mroda88:
We would like to keep the update of the parameters from version 3.2, so this is rejected. It is recommended that before updating the parameters, we also make sure that the code structure is not allowing ambiguities.

PR #24, opened 2019-02-27
Fix memory leaks

igor144:
I did not take the liberty to change the interface, only fixed some memory leaks.


mroda88:
Hi Igor, I don't see the initialisation of the pointer as we said. Where did it go?

igor144:
Hi Marco, these pointers are the private members and they are not deleted now, so there is no need to initialize them to zero. I just commented them out, because, as I said, I don't want to change the interface.

PR #33, opened 2019-06-06
Devel namespace

afropapp13:
New namespace for em events and handling of the Q2 threshold for the relevant case


mroda88:
Sent email with details

PR #34, opened 2019-06-11
Devel namespace

afropapp13:
moved Q2 threshold for em scattering events in relevant namespace under KineUtils


mroda88:
no, wrong definition of the new limits

PR #35, opened 2019-06-11
Devel namespace

afropapp13:
Moving Q2 threshold for em scattering events under the relevant namespace in KineUtils


mroda88:
Afro, please make sure that the documentation is on the docdb

PR #36, opened 2019-06-26
merging from GENIE master branch

afropapp13:
Fixing bug in J for the Rosenbluth cross-section when converting from dSigma/dOmega --> dSigma/dQ2


mroda88:
I have no idea of how you created the pull request. But the only change that was supposed to be there is not there.

PR #47, opened 2019-07-18
Devel cfg final

afropapp13:
Adding the Correlated Fermi Gas Model / high momentum tail in LocalFG.{cxx,h,xml}


mroda88:
Reviewed today, only minor changes expected, then it will be accepted

PR #51, opened 2019-07-29
add Tune name to NtpMCTreeHeader; fill via NtpWriter

nusense:


nusense:
NtpMCTreeHeader didn't have "+" in LinkDef.h so automatic schema evolution isn't all it should be.  Placing the new data member in the middle breaks reading old files (mucks up "runnu" and "daytime").  Cancelling this request; next one will maximize compatibility (as much as possible, but still gives warning message).

PR #52, opened 2019-07-30
Tunename in ntpheader

nusense:
add Tune name to NtpMCTreeHeader; fill via NtpWriter

Improve schema evolution and backwards compatibility.   Still gives warning reading old files, but the data that _is_ there seems to be read fine.


nusense:
Rejecting this pull request in favor of update proposed by Marco.

PR #53, opened 2019-08-21
fix incorrect value of "AGKY-KNO-Alpha-vp"

chenel:
It appears there's a transcription error in the value of the hadronization
parameter "AGKY-KNO-Alpha-vp":
In the work from which the values were drawn, Phys. Rev., D27, 47,
that parameter has value 0.05, while the GENIE value is 0.40.
(All the other parameters agree with the original work,
and there's no indication anywhere that GENIE tuned these values.)
I revised the value to correspond to the PRD one.


chenel:
Per discussion on GENIE Slack, these values will be superseded by a re-tune of the whole parameter set sometime in the (near?) future.  But I think it still makes sense to set the record straight until then.

PR #56, opened 2019-08-30
Ksutton ncgamma aug19

mroda88:
The development of a NC Coherent Gamma process required a number of changes in the framework to allow the coexistence of different types of COH processes at the same time. 
These changes were grouped in this dedicated pull request that could go directly into master as there are no changes in the physics. 
Changes includes hooks for the COH Rho process as well that is into development. 

Tests have been made on the COH PION production model to see if the limits for COH pion were unchanged as a result of the changes in the framework. All ok, See attached file
[gntp.nu_coh.gst.sample_test.pdf](https://github.com/GENIE-MC/Generator/files/3559345/gntp.nu_coh.gst.sample_test.pdf)

I strongly suggest that these changes should be reviews ASAP and merged into the master so the the physics development could go on having a stable framework as starting point.

### DIFF
--- a/config/COHInteractionListGenerator.xml
+++ b/config/COHInteractionListGenerator.xml
@@ -11,17 +11,17 @@ Name             Type     Optional   Comment               Default
 .......................................................................................................
   -->
 
-  <param_set name="CC-PION-Default"> 
+  <param_set name="CC-PION"> 
     <param type="bool" name="is-CC"> true  </param>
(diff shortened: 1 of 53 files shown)


mroda88:
All requested changes from the review with @candreop are implemented and tests have been completed successfully: Physics is unchanged. Plots available in [docdb 144](https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=144)

PR #63, opened 2019-10-07
Short Range Correlation refactorisation 

afropapp13:
Introduction of new module "SRCNuclearRecoil" that handles the possibility of adding of a nuclear recoil and its kinematics.


mroda88:
The new module could go somewhere else instead of NulcearState, in fact it is not a nuclear state, strictly speaking. Hard to debate where.

mroda88:
This pull request will be closed and a new one will be open. The content is the same more or less but the Fermi Momentum code is not duplicated.

PR #64, opened 2019-10-07
Major Updates to the BDM Module

jberger7:
This is a resubmission of a previous pull request.  I updated to a recent (10/04/2019) version of the GENIE master.  A full summary of the current version can be found in https://arxiv.org/abs/1812.05616.  A brief summary includes:

- Inclusion of scattering off electrons;

- Broader set of particle physics models, including both vector and axial couplings, as well as different isospin structures;

- Inclusion of anti-dark matter scattering;

- Improved modeling of the elastic scattering process, including a pseudoscalar form factor;

- Separation of the structure functions from the neutrino structure function code;

- Cross-checked formulae for the cross-sections;


mroda88:
There is clearly some sort of conflict with the commit and so on. We should work on a new pull request that is more organised

src/Framework/Conventions/KinePhaseSpace.h
```diff
@@ -64,6 +64,7 @@ typedef enum EKinePhaseSpace {
   kPSTkTlctl,
   kPSQ2vfE,
   kPSQELEvGen,// Phase space used by genie::QELEventGenerator for sampling kinematic variables
+  kPSDMELEvGen,// Equivalent for Dark Matter scattering
               // TODO: rename this value when the correct variables are identified
   kPSTAfE
 } KinePhaseSpace_t;
```
candreop: Does that enumeration need to go at the bottom for backward compatibility? (at least, till we break connection with GENIE3 for good in GENIE4?) Need comment from @nusense

src/Physics/BoostedDarkMatter/EventGen/DMEKinematicsGenerator.cxx
```diff
@@ -0,0 +1,256 @@
...
+ Important revisions after version 2.0.0 :
+ @ Feb 09, 2009 - CA
+   Moved into the NuE package from its previous location (EVGModules package)
+ @ Feb 06, 2013 - CA
+   When the value of the differential cross-section for the selected kinematics
+   is set to the event, set the corresponding KinePhaseSpace_t value too.
+
+*/
+//____________________________________________________________________________
+
```
candreop: Some large amounts of duplication between the neutrino and DM event generation code - like in this case. While this doesn't prevent accepting the pull request and merge, could we look at reusing (rather than re-implementing) in a future iteration? The code that handles the event generation (for example for generation of event kinematics) is mostly "mechanics", and the physics is  (or should be) contained within the associated cross-section model. This will make it much easier to continue maintaining the BDM module in future releases, and to make sure that BDM benefits from improvements installed in the generic generation modules).

src/Framework/Conventions/KinePhaseSpace.h
nusense: Yes, all additions to enums should go at the end of the list lest one introduce a skew between files written with an earlier version (output is stored as a _value_ not an _enum tag_) but are subsequently read in by a later version.   Enums are great for _code_ but lacking in how the map to values in i/o.

nusense:
The addition of an enum **must** go at the end of the list to preserve correct mapping of names to their values across versions of the code

> [ an enum is ] used to assign names to integral constants, the names make a program easy to read and maintain.

ikatza:
So what's the status of this PR? Will it get merged any time soon?

jberger7:
I fixed the enum ordering issue.  I also merged in some recent bug fixes.  I hope this doesn't create too much of a headache here.  As for the duplication issue, I agree it would be good to avoid this in the future.  I will think on ways to reduce the amount of duplication.  If there are particular tests of the code that would be helpful, I can also contribute that, if you would like.

mroda88:
I did the merge manually as there were some conflicts still.

PR #72, opened 2019-11-12
P8 refact

mroda88:
This is the refactorization of the code so that the common part of Pythia 8 and 6 are not duplicated but they are shared.  
Plots for validation are in [DocDB 163](https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=163)


candreop:
Checked

PR #76, opened 2019-11-21
Add the HEDIS package. Modify GLRES package.

pochoarus:
Addition of the HEDIS module to compute Structure Functions at NLO (requires APFEL). Major modifications of the GLRES package.


src/Framework/Conventions/KinePhaseSpace.h
```diff
@@ -38,6 +38,8 @@ typedef enum EKinePhaseSpace {
   kPSlogyfEx,
   kPSxyfE,
   kPSlogxlogyfE,
+  kPSxQ2fE,
```
mroda88: These should be added at the end to avoid compatibility issues

pochoarus: Ok. Easy to fix.

config/GHE19_00a/CommonParam.xml
```diff
@@ -0,0 +1,53 @@
...
+    <param type="int"    name="NGridX">                                    400 </param>
+    <param type="int"    name="NGridQ2">                                   400 </param>
+    <param type="double" name="XGrid-Min">                                1e-9 </param>
+    <param type="double" name="Q2Grid-Min">                             2.6896 </param>
+    <param type="double" name="Q2Grid-Max">                               1e10 </param>
+    <param type="double" name="MassW">                                  80.385 </param>
+    <param type="double" name="MassZ">                                 91.1876 </param>
+    <param type="double" name="Rho">                                0.00940161 </param>
+    <param type="double" name="Sin2ThW">                                    0. </param>
+    <param type="double" name="CKM-Vud">                               0.97427 </param>
```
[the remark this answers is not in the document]
pochoarus: Ok. I will change it CKM-Vud -> Mud

config/HEDISKinematicsGenerator.xml
```diff
@@ -0,0 +1,38 @@
...
+<!--
+Configuration for the HEDISKinematicsGenerator EventRecordVisitorI
+
+Algorithm Configurable Parameters:
+.......................................................................................................................
+Name                     Type    Opt   Comment                                        Default
+.......................................................................................................................
+MaxXSec-SafetyFactor     double  Yes   multiplies max xsec in rejection method        1.2
+-->
+
```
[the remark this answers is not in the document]
pochoarus: I will merge them in Default.

config/HEDISPXSec.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<!--
+Configuration for the HEDIS d^xsec / dy dx xsec algorithm
+-->
+
+<alg_conf>
+
+  <param_set name="Default"> 
+     <param type="string" name="CommonParam">                    HEDIS-SF  </param>
```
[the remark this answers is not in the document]
pochoarus: Some of the parameters that are the list of CommonParam-HEDIS are used by other algorithms. For instance, XGrid-Min is used by HEDISXSec, HEDISPXSec and HEDISKinematicsGenerator. That is why a created a CommonParam file for HEDIS. How could I use the parameters defined in HEDISPXSec.xml in HEDISXSec or HEDISKinematicsGenerator?

src/Physics/GlashowResonance/EventGen/GLRESGenerator.cxx
```diff
@@ -58,83 +68,197 @@ GLRESGenerator::~GLRESGenerator()
...
+  
+    // Compute the neutrino and muon energy
+    double Ev  = (long double) init_state.ProbeE(kRfLab); 
+    double El  = y*Ev;
+
+    LOG("GLRESGenerator", pINFO) << "Ev = " << Ev << ", y = " << y << ", -> El = " << El;
+    
+    // Compute the momentum transfer and scattering angle
+    double El2   = TMath::Power(El,2);
+    double me    = (long double) kElectronMass;
```
mroda88: these casts don't make any sense. What were you trying to do?

src/Physics/GlashowResonance/EventGen/GLRESGenerator.cxx
```diff
@@ -58,83 +68,197 @@ GLRESGenerator::~GLRESGenerator()
...
+  LOG("GLRESGenerator", pINFO) << "Channel : " << interaction->FSPrimLeptonPdg();
+
+  if ( pdg::IsElectron(pdgl) || pdg::IsMuon(pdgl) || pdg::IsTau(pdgl) ) {
+
+    // Get selected kinematics
+    double y = interaction->Kine().y(true);
+    assert(y>0 && y<1);
+  
+    // Compute the neutrino and muon energy
+    double Ev  = (long double) init_state.ProbeE(kRfLab);
```
mroda88: Also this cast is nonsense

src/Physics/GlashowResonance/XSection/GLRESPXSec.cxx
```diff
@@ -94,9 +164,27 @@ bool GLRESPXSec::ValidProcess(const Interaction* interaction) const
...
 }
 //____________________________________________________________________________
+void GLRESPXSec::LoadConfig(void)
+{
+
+  //-- load the differential cross section integrator
+  fXSecIntegrator = dynamic_cast<const XSecIntegratorI *> (this->SubAlg("XSec-Integrator"));
+  assert(fXSecIntegrator);
+
+  GetParam( "Wmin", fWmin ) ;
```
mroda88: Can we make the change of this Wmin parameter more specific? Suggestion is GLRES-Wmin we have other Wmin around the code and if would be much better if we could distinguish them easily

src/Physics/HEDIS/XSection/HEDISStrucFunc.h
```diff
@@ -0,0 +1,267 @@
...
+    double Vud,Vus,Vub;
+    double Vcd,Vcs,Vcb;
+    double Vtd,Vts,Vtb;
+
+  };
+
+  inline bool operator== (const SF_info& a, const SF_info& b) {
+    
+    bool ok = true;
+    if ( a.LHAPDFmember != b.LHAPDFmember   ) { ok = false; std::cout << "LHAPDFmember" << std::endl; }
```
mroda88: We don't usually use cout or cerr in GENIE

src/Physics/HEDIS/XSection/HEDISStrucFunc.h
```diff
@@ -0,0 +1,267 @@
...
+        kSFT3, 
+        kSFnumber, 
+      } 
+      HEDISStrucFuncType_t;
+
+      // ................................................................
+      // HEDIS form factor type
+      //
+
+      class HEDISStrucFuncTable
```
mroda88: Do we need a singleton for this?

mroda88: Another way to express my doubts here is: why can we not have in the memory two different tables?

src/Physics/Hadronization/LeptoHadronization.h
```diff
@@ -0,0 +1,68 @@
+//____________________________________________________________________________
+/*!
+
+\class    genie::LeptoHadronization
+
+\brief    Provides access to the LEPTO hadronization models. \n
+          Is a concrete implementation of the HadronizationModelI interface.
```
mroda88: Strictly speaking this does not exist anymore

src/Physics/Hadronization/LeptoHadronization.h
```diff
@@ -0,0 +1,68 @@
...
+\cpright  Copyright (c) 2003-2018, The GENIE Collaboration
+          For the full text of the license visit http://copyright.genie-mc.org
+          or see $GENIE/LICENSE
+*/
+//____________________________________________________________________________
+
+#ifndef _LEPTO_HADRONIZATION_H_
+#define _LEPTO_HADRONIZATION_H_
+
+#include <TPythia6.h>
```
mroda88: This should be inside some `#ifdef` in the case when pythia 6 will be removed

src/Physics/Hadronization/LeptoHadronization.cxx
```diff
@@ -0,0 +1,515 @@
...
+  // Width of Gaussian distribution for transverse momentums
+  // Define in LEPTO with PARL(3) and PARL(14)
+  GetParam("Primordial-kT", fPrimordialKT ) ;
+  GetParam("Remnant-pT",    fRemnantPT ) ;
+
+  // It is, with quark masses added, used to define the minimum allowable energy of a colour-singlet parton system.
+  GetParam( "Energy-Singlet", fMinESinglet ) ; 
+
+  // PYTHIA parameters only valid for HEDIS
+  GetParam( "Wmin", fWmin ) ;
```
mroda88: Same comment on Wmin than in the GLRES. Can we make its name more specific? Is it the same quantity for GLRES? is it a different one?

src/Physics/XSectionIntegration/GSLXSecFunc.cxx
```diff
@@ -187,6 +187,46 @@ ROOT::Math::IBaseFunctionMultiDim *
...
+}
+double genie::utils::gsl::d2XSec_dlog10xdlog10Q2_E::DoEval(const double * xin) const
+{
+// inputs:  
+//  log10(x) [-]
+// log10(Q2) [-]
+// outputs: 
+//   differential cross section [10^-38 cm^2]
+//
+  double  log10x = xin[0];
```
mroda88: I appreciate the clarity of the passage here, but this function is supposed to be called a lot of times, can we avoid memory duplication here? 
Just pass xin[0] and xin[1] at the KinePtr setters

src/make/Make.include
```diff
@@ -492,6 +506,7 @@ CPP_INCLUDES := \
     $(ROOT_INCLUDES) \
     $(PYTHIA8_INCLUDES) \
     $(LHAPDF_INCLUDES) \
+    $(APFEL_INCLUDES) \
```
mroda88: I suggest to include the APFEL libraries only when required in the specific makefiles. not at the top level. Examples can be seen in [Robet's code](https://github.com/nusense/Generator/blob/inclxx_and_g4/src/Physics/HadronTransport/Makefile) where we include INCL++ and Geant4 as specific packages

candreop:
That looks really very good to me Alfonso, with a small caveat (see below), and should be ready to go for v3.2. As Marco said, we can help you address any residual if we can all agree / manage to coordinate. I think what is missing is a greater degree of uniformity with the rest of GENIE - There is really no need for special "HEDIS" labelling (eg HEDISQrkChannel_t in XclsTag): We do have DIS labels and, separately, we provide the energy of the event (whether that is high evergy, very high energy, ultra high energy etc is a subjective definition and, in any case, we should avoid contaminating labeling of interaction types with language that refers to modelling choices. Some of the things added (extra variables for hit quarks etc, as well as HEDISQrkChannel_t, could be made part of standard DIS too). If, for example, we didn't need special HEDIS labels in Interaction and a special HEDIS InteractionListGenerator etc (because the standard DIS ones could be reused), and if HEDISQrkChannel_t was actually not needed because the quark PDGs are there in XclsTag, we could remove 1-2 thousand lines of almost redundant code.

candreop:
For the post-v3.2 development: We've already brought in PYTHIA8 as an alterative hadronization model, and we're now looking to change the few residual places in the code where PYTHIA6 is still requied (charm hadronization and decays). Would you be able in the future to provide an alternative version of LeptoHadronization that depends on PYTHIA8 (rather than 6)?

pochoarus:
OK, I can try to homogenize the HEDIS with respect to DIS. In addition, I will include most of the changes suggested by Marco.

On the other hand, there is one remain issue related to the CommonParameters.xml file. There are some parameters that are used in different modules (xsec, kinematics generation, hadronization). Currently, I only know how to use them by defining them in the CommonParameters.xml. Is there another way to handle it?

config/EventGeneratorListAssembler.xml
```diff
@@ -58,16 +58,42 @@ Generator-%d                alg     No
...
-     <param type="int" name="NGenerators">  4                                  </param>
-     <param type="alg" name="Generator-0">  genie::EventGenerator/GLRES        </param>
-     <param type="alg" name="Generator-1">  genie::EventGenerator/DIS-CC       </param>
-     <param type="alg" name="Generator-2">  genie::EventGenerator/DIS-NC       </param>
-     <param type="alg" name="Generator-3">  genie::EventGenerator/DIS-CC-CHARM </param>
+     <param type="int" name="NGenerators">  7                                  </param>
+     <param type="alg" name="Generator-0">  genie::EventGenerator/GLRES-Mu     </param>
+     <param type="alg" name="Generator-1">  genie::EventGenerator/GLRES-Tau    </param>
+     <param type="alg" name="Generator-2">  genie::EventGenerator/GLRES-Ele    </param>
+     <param type="alg" name="Generator-3">  genie::EventGenerator/GLRES-Ele    </param>
```
hrgallagher: Why is GLRES-Ele in the list twice?

src/Framework/Utils/KineUtils.cxx
```diff
@@ -398,8 +407,8 @@ Range1D_t genie::utils::kinematics::InelXLim(double Ev, double M, double ml)
   assert (s>M2);
 
   Range1D_t x;
-  x.min = ml2/(s-M2) + controls::kASmallNum;
-  x.max = 1.         - controls::kASmallNum;
+  x.min = ml2/(s-M2) + controls::kAVerySmallNum;
```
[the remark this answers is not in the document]
hrgallagher: I don't think the difference will have a noticeable effect for other GENIE processes, and if it does, the new code is more correct than the old x_min calculation, so I'm fine with it.

src/Physics/GlashowResonance/XSection/GLRESPXSec.cxx
```diff
@@ -46,32 +51,97 @@ GLRESPXSec::~GLRESPXSec()
...
+
+  double y     = kinematics.y();
+  double s0    = 2 * kElectronMass * E;
+
+  double Gw      = PDGLibrary::Instance()->Find(kPdgWM)->Width(); //PDGLibrary::Instance()->Find(kPdgWM)->Width() //genhen=2.124
+  double gf      = kGF2/kPi;                                      //kGF2/kPi                                      //genhen=1.16637e-05*1.16637e-05/3.141592653589793
+  double m_w     = kMw;                                           //kMw                                           //genhen=80.425
+  double m_z     = kMz;                                           //kMz                                           //genhen=91.1876
+  double Sin2thw = 1 - kMw2 / kMz2;                               //1 - kMw2 / kMz2                               //genhen=0.2312
+  double branch  = 64.41/10.63;                                   //64.41/10.63                                   //genhen=67.96/10.57
```
hrgallagher: can you add a comment here explaining what branch is?

pochoarus:
I have pushed most of the requests from the review into dev_hedis.

The most important updates:
- Now we dont distinguish between HEDIS or DeepInelastic channels, so everything is more homogeneus.
-LeptoHadronization includes some #ifdef in case pythia6 is not enable.
-APFEL flags are only loaded in HEDIS/Xsec Makefile. 

There is only the pending issue, from Marco, with the CommonParameters.xml files that are available three HEDIS tunes.

pochoarus:
I have added the new treatment to decay unstable particles in pythia6 using CommonDecay.xml files. I have created an individual CommonDecay.xml for the three tunes we currently have GHE19_00a, GHE19_00b, GHE19_00c. This config file is different to the global CommonDecay.xml file (different particles are decayed).

mroda88:
HI Alfonso, I added my changes in configuration and in the gSpline2Root. And for me I think it's ok. @candreop what do you think?

PR #78, opened 2019-12-02
Hadro rename

mroda88:


mroda88:
The bux fix is about the production of charmed mesons. The plots I attach are the PDG of particle vs the energy of the neutrino. In the Old master the production of D (PDG ~ 400) are stopping after 100 GeV. The new version has charmed mesons up to 200 GeV, which is the maximum of the production. 
[Fixed version](https://github.com/GENIE-MC/Generator/files/3941013/charm_pdg_new.pdf)
[old master](https://github.com/GENIE-MC/Generator/files/3941020/charm_pdg_old.pdf)

PR #81, opened 2019-12-24
Do Not Merge - Review of the implementation of Minoo's single pion model

mroda88:


igor144:
Hi all,

Please, let's postpone the problem with parameters, which, of course, is very important but maybe not the most urgent. We guess the first and most important issue is related to problems and bugs outlined in our recent report
<https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=181>, since, without their solving (even voluntaristic) one cannot move forward to the implementation of simulation (not to rewrite the code again and again).

The second, maybe terminological remark is related to some misunderstanding. Let's draw your attention that ALL MODELS: KLN, BS, R (not in GENIE), and MK ARE ESSENTIALLY THE REIN-SEHGAL BASED MODELS. ALL THESE MODELS are RESONANCE SINGLE PION PRODUCTION models and all take into account (one or another way) the non-resonance background (NRB) contributions.
Rein's and MK models differ from the RS, KLN, and BS in the NRB and in more detailed pion production dynamics (4d cross sections instead of 2d ones). Other details (form factors, phase factors) are quite "technical". To clarify this point, please, look at the diagram on page 2 of the mentioned report, where we show the genesis of these models. Thus, it seems there is no need
in a dedicated folder for the SPP model(s), or otherwise one has to put in that folder all aforementioned models.
Even if we decide to have the "SPP-model" folder, the MK model deserves its own name, like all other resonance models have their own names in the folder Physics/Resonance.

mroda88:
Thanks for your explanations! I'll try to have a look at the rest of your code ASAP. And sure, I agree we should focus on the model not on the parameters, I was pointing out that should be a project on its own.

PR #87, opened 2020-02-18
Fix dipole form factor in the Empirical MEC model

afropapp13:
Fixing the issue with the exponent in the dipole form factor squared in the Empirical MEC model


mroda88:
This pull request is not much relevant anymore given the soon inclusion of SuSAv2

PR #90, opened 2020-02-18
Add the NCinclMEC event generator list

sjgardiner:
GENIE v3.0.6 lacks a standard event generator list for NC inclusive that includes MEC. I've added a new one called NCinclMEC (based on CCinclMEC) to address this.

### DIFF
--- a/config/EventGeneratorListAssembler.xml
+++ b/config/EventGeneratorListAssembler.xml
@@ -476,6 +476,14 @@ Generator-%d                alg     No
      <param type="alg" name="Generator-2">   genie::EventGenerator/DME      </param>
   </param_set>
 
+  <param_set name="NCinclMEC">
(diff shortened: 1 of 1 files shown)


sjgardiner:
My preference would be to update both to include DFR, but I could be persuaded otherwise. What do you think, @candreop? I can make that change if there is agreement.

candreop:
The CCInclMEC (NCInclMEC) were first produced when MEC was first discussed about. Back then the MEC implementation in GENIE was very rough and the CCInclMEC was meant to be a non-standard generator list used for studying the likely impact of MEC on analyses. Now MEC is a standard in all tunes and we have several well-motivated models. How about we drop the *InclMEC generator lists and just update (add MEC) in the standard CC and NC?

mroda88:
I absolutely like the solution suggested by @candreop . @sjgardiner does this meet your needs ?

PR #93, opened 2020-02-19
automatically select flux surface radii for ROOT geometries

tlatorre-uchicago:
This commit updates gevgen_atmo to automatically select the longitudinal and
transverse radii for the flux surface when using a ROOT geometry and if the two
radii are not specified on the command line. If this is the case, both are set
to the diagonal of the bounding box of the geometry.

I *believe* the diagonals of the bounding box should be large enough to correctly sample the neutrino interaction vertices assuming that the origin of the ROOT geometry is within the bounding box.

I have *not* extensively tested this. So far I have done some basic testing with a similar commit applied to version 2.12.8 and have just made sure that the pull request compiles on master so it should probably get tested more before being merged.

### DIFF
--- a/src/Apps/gAtmoEvGen.cxx
+++ b/src/Apps/gAtmoEvGen.cxx
@@ -233,8 +233,12 @@
 #include <vector>
 #include <sstream>
 #include <map>
+#include <iomanip>
 
 #include <TRotation.h>
+#include <TMath.h>
+#include <TGeoShape.h>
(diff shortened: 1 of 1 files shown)


candreop:
This seems fine to me. It is a clever trick. Any comment @nusense ? Otherwise will merge the pull request

nusense:
Sorry, I overlooked the request for comment here.  No this look good.  I love it when best guesses get handled automagically.

PR #94, opened 2020-02-19
abort with an error if LoadFluxData() fails

tlatorre-uchicago:
This commit updates gevgen_atmo to abort with an error if one of the flux files
fails to load when calling LoadFluxData(). I got bit by this where I
accidentally mistyped a single flux file and that flux failed to load, but
there was only a small warning but everything else continued to run.

I also updated LoadFluxData() to immediately abort if a single flux
file fails to load instead of continuing in order to make sure that the error
message about the file failing to load comes last.

I also fixed a small typo in gevgen_atmo.

I have tested this with a similar commit to version 2.12.8 but have only build tested the current pull request on master.


candreop:
Thanks Anthony

PR #98, opened 2020-03-26
added 'gen-unscaled' option to gevgen

marialiubarska:
If set, will overwrite Weight with Weight*GlobalProbabilityScale, so files produced with different energy ranges can be combined. This is needed for IceCube GENIE reader to work correctly (events are split into few energy ranges during production).
If not set, nothing will be changed.

### DIFF
--- a/src/Apps/gEvGen.cxx
+++ b/src/Apps/gEvGen.cxx
@@ -36,6 +36,7 @@
                   [--cache-file root_file]
                   [--xml-path config_xml_dir]
                   [--tune G18_02a_00_000] (or your preferred tune identifier)
+                  [--gen-unscaled]
 
          Options :
            [] Denotes an optional argument.
@@ -108,6 +109,9 @@
               re-used in subsequent MC jobs.
            --xml-path
               A directory to load XML files from - overrides $GXMLPATH, and $GENIE/config
+	   --gen-unscaled
+	      Write unscaled weight option (required to correcly combine 
+	      files generated with different E ranges/geometries)
 
         ***  See the User Manual for more details and examples. ***
 
@@ -118,7 +122,7 @@
 
 \cpright Copyright (c) 2003-2020, The GENIE Collaboration
          For the full text of the license visit http://copyright.genie-mc.org
-
+         
 */
 //____________________________________________________________________________
 
@@ -214,6 +218,7 @@ long int        gOptRanSeed;      // random number seed
 string          gOptInpXSecFile;  // cross-section splines
 string          gOptOutFileName;  // Optional outfile name
 string          gOptStatFileName; // Status file name, set if gOptOutFileName was set.
+bool            gOptGenUnscaled = false;  // write EventRecorg->Weight() in global probability scale
 
 //____________________________________________________________________________
 int main(int argc, char ** argv)
@@ -312,7 +317,7 @@ void GenerateEventsAtFixedInitState(void)
 
      // generate a single event
      EventRecord * event = evg_driver.GenerateEvent(nu_p4);
-
+     
      if(!event) {
         LOG("gevgen", pNOTICE)
           << "Last attempt failed. Re-trying....";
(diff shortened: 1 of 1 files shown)


candreop:
Hi Maria. 
Thank you for your contribution. Looking at this briefly, it seems to me that the weight of individual events (which is kind of reserved for physics modelling weights) is not the right place to store overall normalisation factors for an entire sample. You may want to look at some of the apps that are specialised for different experiments (eg. gT2KEvGen.cxx in Apps/). There, you will see this line: **ntpw.EventTree()->SetWeight(pot);**
which sets the overall sample normalisation once (on the event tree) and not on each event.
Would something like this work for you?

marialiubarska:
Hi Costas, thank you very much for your reply! 
I wasn't aware of this option and it indeed makes more sense to set weight for the whole tree. This will work for me, so I updated the code.
Please, let me know if you have other comments or suggestions.

candreop:
Hi Maria. OK, thanks for updating the code to store an overall weight for the entire sample. Now, perhaps we can focus on what is it that we actually store there (given that this is a generic app with options that should be meaningful in all contexts and it doesn't implement experiment-specific conventions etc. We have experiment-specific event generation apps for this purpose). The best option is to store the absolute normalisation of each sample (as it is done in the experiment-specific apps), but this is somewhat difficult to do generically for all situations. It probably makes sense to try and add a weight expressing the relative difference in exposure between different types of event samples, so that they can be combined more easily, leaving a single absolute normalisation factor to be computed/applied externally. What do you propose to store. I am including @nusense in the discussion too.

nusense:
I do think making use of the tree weight definitely is the right option.  I'm not sure that for plain-old `gevgen` that the weight could ever mean much in particular (other than relativeness to other similar samples).  We make no requirement on normalization of the input fluxes that `gevgen` supports, so it's hard to assign any particular units to what's recorded.

That means it kind of has to be up to the user to make sensible choices when combining sets.  And I'm not sure what more we can do.   Did you have some particular idea Costas?

marialiubarska:
Hi!
I would say that having `GlobProbScale` saved or included in event weight would be the easiest option for me.

In the past, there was a simple GENIE app implemented inside IceCube software where `GlobProbScale` was manually saved. It then was used to set event normalization to 1 and, because of that, it was part of event weight used later in the analysis. 
At the moment, I'm not entirely sure that having constant normalization for all events would necessarily be a problem and I'm currently trying to confirm this with more knowledgeable people, but having `GlobProbScale` stored will definitely be the safest and easiest option.

We were thinking about having IceCube specific app, but, since in the current simulation setup we do not need complex geometry and flux, the only additional thing we need is normalization (saved separately or included in event weight). So I thought that it makes more sense to just add the corresponding option to `gevgen` (which might be also useful for other people who need to combine different samples).

marialiubarska:
I was told that, other than for combining files with different energy ranges, there are some cases when we want to know general normalization. So it would be great if there is a way to get it from the output files.

candreop:
It seems to me Maria and @nusense that the best thing to do would be to go as close as possible to what is done by the experiment-specific apps that compute and write out the absolute normalisation of the event sample. This is not easy to do in the generic event generation app (gevgen) because a) it uses a simple`target mix' (not an actual detector geometry) and it doesn't need to know about the detector mass, and b) because the input flux format for this app (either histogram, tabular text file, or some functional form) does not specify the absolute scale / units of the flux, as different options are best for different contexts. But, within gevgen, we can calculate the appropriate `flux * xsection * density/A' integral over the energy range of the input flux, and incorporate the internal `probability scale', so what we write out is only one (user-specific) multiplication constant away from the correct absolute normalisation for the generated sample. What do you think?

marialiubarska:
Hi Costas!
I'm not completely sure if I understood this correctly. Do you mean that, after calculating integral `Int = integral flux*xsec*density/A dE` for a given energy range,  the sample is renormalized inside `gevgen` by `GlobProbScale/Int` and then tree weight is set to `Int`?
If yes, this will be fine for me.

PR #101, opened 2020-04-06
fix header file includes in src/contrib/test

tlatorre-uchicago:
This commit is an attempt to get the test files in src/contrib/test to compile
correctly. Mostly this has involved updating the header file includes at the
start of each file since they were reorganized at some point. I also had to
update a handful of tests to change names here and there to get things to
compile.

Currently everything compiles except for two tests:

1. gtestDecay.cxx

In this test, there are a number of calls to Decayer::InhibitDecay() which
appears to now be a protected member function, so I'm not sure what the best
way to update this is.

2. gtestHadronization.cxx

This test relies on a file which used to be present called
HadronizationModelI.h, but which was removed in commit
9dec82dec024367dcc4253fcc6a6d3b4151c0803. Since there is no commit message for
that commit, I don't have enough knowledge to figure out how to fix this.


candreop:
Thanks! The gtestDecay.cxx and gtestHadronization.cxx are more difficult to update since, recently, decay and hadronization model interfaces changed: They became 'event record visitors' themselves (rather than external algorithms called by event record visitors) and they operate on full events (GHEP records). Will see what is possible to do and if is useful to maintain these simple tests. But what you've done already is very useful!

PR #102, opened 2020-05-01
Ensure that the primary lepton polarization is set by all relevant generator classes

sjgardiner:
Move code to set the primary lepton polarization to a utility function. Set this polarization in the MECGenerator, QELEventGenerator, and QELEventGeneratorSM event record visitors.


candreop:
Looks good to me Steven. Thanks

PR #103, opened 2020-05-04
Event library interface

cjbackhouse:
for review meeting Thurs 7 May


src/Tools/VMC/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,261 @@
...
+}
+//____________________________________________________________________________
+void EventLibraryInterface::ProcessEventRecord(GHepRecord * event) const
+{
+// Get event summary constructed by GENIE
+//
+  Interaction* interaction = event->Summary();
+  const InitialState & init_state = interaction->InitState();
+
+  const Record* rec = GetRecord(interaction);
```
nusense, line 54: The use of generic class name `Record` here within `ProcessEventRecord(GHepRecord * event)` had me do a double-take.  Could it be changed up to something like `EvtLibRecord` ?

src/Tools/VMC/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+  if(!proc.IsWeakCC() && !proc.IsWeakNC()){
+    LOG("ELI", pINFO) << "Skipping unknown process " << proc;
+    return 0;
+  }
+
+  int probe_pdgc = init_state.ProbePdg();
+
+  // Use nu_mu for NC by convention
+  if(proc.IsWeakNC()){
```
nusense, line 141: This convention of tagging all the NC's a nu_mu generated is something that one has to impose when generating the event library, then?

src/Tools/VMC/EvtLibInteractionListGenerator.cxx
```diff
@@ -0,0 +1,62 @@
...
+InteractionList * EvtLibInteractionListGenerator::CreateInteractionList(
+   const InitialState & init_state) const
+{
+  InteractionList * intlist = new InteractionList;
+
+  // REVIEW these are all these arbitrary choices, but seems to be required to
+  // be able to succesfully pick a spline. Perhaps it would be better to add
+  // new enum entries for "unspecified because it came from an external
+  // library"?
+
```
nusense, line 39: Yes, rather thane using `kScQuasiEleastic` we should add a new "scattering type" for external generators.  In my parallel start on something similar (but completely imcomplete), I added ` kScExternalGen = 999`, see
https://github.com/nusense/Generator/blob/extgen/src/Framework/Interaction/ScatteringType.h#L55

src/Tools/VMC/Utils.cxx
```diff
@@ -0,0 +1,34 @@
+#include "Tools/VMC/Utils.h"
+
+#include "Framework/Messenger/Messenger.h"
+
+#include <wordexp.h>
+
+//___________________________________________________________________________
+// REVIEW is there a built-in tool for doing this?
```
nusense, line 8: Not exactly -- in that none use `wordexp` to do all of:
>  The expansion done consists of the following stages: tilde expansion
       (replacing ~user by user's home directory), variable substitution
       (replacing $FOO by the value of the environment variable FOO),
       command substitution (replacing $(command) or `command` by the output
       of command), arithmetic expansion, field splitting, wildcard
       expansion, quote removal.
I'm not so familiar with `wordexp` ... if wildcard expansion yields more than one file, what's the result?  And can your code handle that?

Makefile
```diff
@@ -202,6 +203,18 @@ else
 	@echo "** Building geometry-drivers was not enabled. Skipping..."
 endif
 
+tools-vmc: FORCE
```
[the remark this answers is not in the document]
cjbackhouse: This was the name in Costas' original template. I'm happy to change it to whatever is the consensus.

config/G18_02a/ModelConfiguration.xml
```diff
@@ -138,7 +138,7 @@ STFC, Rutherford Appleton Laboratory
   <param type="alg" name="XSecModel@genie::EventGenerator/NucleonDecay">   genie::DummyPXSec/Default             </param>
   <param type="alg" name="XSecModel@genie::EventGenerator/NNBarOsc">       genie::NNBarOscDummyPXSec/Default     </param>
 
-
+  <param type="alg" name="XSecModel@genie::EventGenerator/EventLibraryInterface">  genie::vmc::EvtLibPXSec/Default   </param>
```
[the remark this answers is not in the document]
cjbackhouse: I admit to being a little lost in all the config files and tunes.
Ideally this would be the "gibuu external v whatever" tune. And then we could have have the "nuwro external v something else" tune, with the same config, just a different library configured. That would presumably prevent the spline xmls getting mixed up.

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: The library has directories like cc/nu_mu, but it also has nc/nu and nc/nu_bar. This NC=numu convention is internal to the code. I also generated them that way, but it doesn't matter. I rewrite the outgoing neutrino flavour elsewhere.

src/Tools/VMC/Utils.cxx
cjbackhouse: wordexp() can handle it, but I abort in that case. L28.

config/EventLibraryInterface.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<alg_conf>
+  <param_set name="Default">
+    <param type="string" name="LibraryPath"> Dummy.root </param>
```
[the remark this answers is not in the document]
cjbackhouse: Can do x2

config/EvtLibPXSec.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<alg_conf>
+  <param_set name="Default">
+    <param type="string" name="LibraryPath"> Dummy.root </param>
```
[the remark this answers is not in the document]
cjbackhouse: Can do

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: Can do (or we could be more explicit with namespaces?)

src/Tools/VMC/Utils.cxx
cjbackhouse: The main point here is so that experiments can specify something like $EVENT_LIBRARY_DIR/lib.root and have that managed by ups or some similar system.

src/Tools/VMC/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,261 @@
...
+{
+// Get event summary constructed by GENIE
+//
+  Interaction* interaction = event->Summary();
+  const InitialState & init_state = interaction->InitState();
+
+  const Record* rec = GetRecord(interaction);
+  if(!rec) return; // Reason has already been printed
+
+  std::unique_ptr<TLorentzVector> probe_p4(init_state.GetProbeP4(kRfLab));
```
mroda88, line 57: This is not ok if we want to release it in v 3.2, otherwise it will be ok

src/Tools/VMC/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+  unsigned int nLep = 0;
+  for(const Particle& part: rec->parts) if(pdg::IsLepton(part.pdg)) ++nLep;
```
mroda88, line 60: same here: not C++98

src/Tools/VMC/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,261 @@
...
+    const std::string& tgtName = dir->GetName();
+    const TParticlePDG* tgtPart = pdglib->DBase()->GetParticle(tgtName.c_str());
+    if(!tgtPart){
+      LOG("ELI", pWARN) << "Unknown nucleus " << tgtName
+                        << " found in " << libPath
+                        << " -- skipping";
+      continue;
+    }
+
+    for(int sign: {+1, -1}){
```
mroda88, line 225: Let's use the constants we defined, not hard coded numbers

src/Tools/VMC/EvtLibPXSec.cxx
```diff
@@ -0,0 +1,154 @@
...
+{
+  Algorithm::Configure(config);
+}
+
+//____________________________________________________________________________
+void EvtLibPXSec::LoadXSecs() const
+{
+  std::string libPath;
+  if(!GetParamDef( "LibraryPath", libPath, std::string() )){
+    LOG("ELI", pFATAL) << "Must specify 'LibraryPath'";
```
mroda88, line 72: Don't decide here how to kill the code. The code already has an exit if the parameter is not found. Just use GetParam( "LibraryPath", libPath ) ; and you'll be fine. 

See comments from Robert about the name of this variable

src/Tools/VMC/EvtLibPXSec.cxx
```diff
@@ -0,0 +1,154 @@
...
+    const std::string& tgtName = dir->GetName();
+    const TParticlePDG* tgtPart = pdglib->DBase()->GetParticle(tgtName.c_str());
+    if(!tgtPart){
+      LOG("ELI", pWARN) << "Unknown nucleus " << tgtName
+                        << " found in " << libPath
+                        << " -- skipping";
+      continue;
+    }
+
+    for(int sign: {+1, -1}){
```
mroda88, line 95: let's use our PDG constants

src/Tools/VMC/EvtLibPXSec.h
```diff
@@ -0,0 +1,40 @@
...
+  // Override the Algorithm::Configure methods to load configuration
+  // data to private data members
+  void Configure (const Registry & config);
+  void Configure (string param_set);
+
+protected:
+  TGraph* GetXSec(const Interaction* in) const;
+  void LoadXSecs() const;
+
+  mutable std::map<Key, TGraph*> fXSecs;
```
mroda88, line 34: why is this mutable?

src/Tools/VMC/Key.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <ostream>
+
+namespace genie{
+namespace vmc{
+
+struct Key
+{
+  Key(int _nucl_pdg, int _nu_pdg, bool _iscc)
+    : nucl_pdg(_nucl_pdg), nu_pdg(_nu_pdg), iscc(_iscc) {}
```
mroda88, line 12: I'm sure we will cover this at the review today, but why only CC and NC? We know this is not enough.

src/Tools/VMC/RecordList.h
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  //---------------------------------------------------------------------------
+  struct Record
+  {
+    Record();
+    Record(float _E, /*float _w,*/ int _prod_id, const std::vector<Particle>& _ps);
+
+    /// Order by energy
+    bool operator<(const Record& rhs) const;
```
mroda88, line 30: This is surprising, do we really need to order this kind of objects?

src/Tools/VMC/RecordList.h
```diff
@@ -0,0 +1,100 @@
...
+
+    long NRecords() const;
+    Record GetRecord(int i) const;
+  protected:
+    TTree* fTree;
+
+    float Enu;//, weight;
+    int prod_id;
+    int nparts;
+    int pdgs[1024];
```
mroda88, line 62: fixed memory? Why cannot we use vector?

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: I had no idea genie was compatible with such ancient standards. I take it the plan is to drop that requirement soon?

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: sorry to be slow, which constants exactly?

src/Tools/VMC/EvtLibPXSec.h
cjbackhouse: Because Integral() has to be declared const, and I call LoadXSecs() lazily there. I suppose in principle one could call it at the end of Configure()? I don't know what the normal coding convention is. My claim is that the current structure is fine - logically the function is const, it's just an implementation detail that it happens to populate a cache.

src/Tools/VMC/RecordList.h
cjbackhouse: This is the target of a SetBranchAddress(). There's no way to know the number of particles up-front. A vector that's fixed to 1024 elements isn't really better. std::array is marginally better, but probably not available in old C++. I think it's fine to just set some extremely high upper limit and trust the event libraries to keep within it.

Makefile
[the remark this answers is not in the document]
cjbackhouse: Is there a specific suggestion here?

src/Tools/VMC/EvtLibPXSec.cxx
cjbackhouse: Done

config/EventLibraryInterface.xml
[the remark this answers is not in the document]
cjbackhouse: Done

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: Done

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: Improved the comments.

src/Tools/VMC/EvtLibPXSec.h
cjbackhouse: I worked around this one.

src/Tools/VMC/RecordList.h
cjbackhouse: Added a comment.

src/Tools/VMC/RecordList.h
cjbackhouse: Added a comment.

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: How do I set my built up to use C++98? There look to be a lot of places and I'm likely to miss one.

Alternately -- is it possible to have this component optional, and require a higher compiler version if it's enabled? I don't think any potential users are on anything like so old a standard.

Makefile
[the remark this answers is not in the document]
cjbackhouse: How about just EvtLib/?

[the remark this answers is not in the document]
cjbackhouse: I went with genie::evtlib and Tools/EvtLib

src/Tools/VMC/EvtLibInteractionListGenerator.cxx
cjbackhouse: Done

src/Tools/VMC/EventLibraryInterface.cxx
cjbackhouse: Done

src/Tools/VMC/EvtLibPXSec.cxx
cjbackhouse: Done

config/G18_02a/ModelConfiguration.xml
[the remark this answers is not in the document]
cjbackhouse: Done, as per discussion on slack.
Use --tune EX00_00a_00_000 to run.

mroda88:
I tested the code. There is still come problem with the definition of daughter/mother relations

cjbackhouse:
Can you be more specific?

mroda88:
```|------------------------------------------------------------------------------------------------------------------|
|GENIE GHEP Event Record [print level:   3]                                                                        |
|------------------------------------------------------------------------------------------------------------------|
| Idx |          Name | Ist |        PDG |   Mother  | Daughter  |      Px |      Py |      Pz |       E |      m  | 
|------------------------------------------------------------------------------------------------------------------|
|   0 |         nu_mu |   0 |         14 |  -1 |  -1 |   2 |   3 |   0.000 |   0.000 |   1.310 |   1.310 |   0.000 | 
|   1 |           C12 |   0 | 1000060120 |  -1 |  -1 |   3 |   4 |   0.000 |   0.000 |   0.000 |  11.175 |  11.175 | 
|   2 |       neutron |  11 |       2112 |  -1 |  -1 |   3 |   3 |   0.000 |   0.000 |   0.000 |   0.940 |   0.940 | 
|   3 |           mu+ |   1 |        -13 |   0 |  -1 |  -1 |  -1 |  -0.062 |   0.530 |   0.919 |   1.068 |   0.106 | 
|   4 |        proton |   1 |       2212 |   1 |  -1 |  -1 |  -1 |   0.167 |  -0.645 |   0.177 |   1.164 |   0.938 | 
|------------------------------------------------------------------------------------------------------------------|
|       Fin-Init:                                                |   0.105 |  -0.115 |  -0.214 | -10.253 |         | 
|------------------------------------------------------------------------------------------------------------------|
```
This is a snippet from one event. I don't know how it could be that the muon is coming out of a neutron.

cjbackhouse:
Am I misreading that? Doesn't it say the mother index is zero, ie the neutrino?

mroda88:
yes, it says that and the daughter of the neutron is the muon as well.

config/EventGenerator.xml
```diff
@@ -115,6 +115,16 @@ XSecModel                   alg     Yes  Cross section model used at the thread
   </param_set>
 
 
+ <!--
+  EvtLib / EventLibraryInterface
+ -->
+  <param_set name="EventLibraryInterface">
+     <param type="string" name="VldContext"> </param>
+     <param type="int"    name="NModules">   1                                                    </param>
```
mroda88, line 123: We need a dedicated inital state appender or the flux driver will never work properly

src/Framework/Interaction/ProcessInfo.h
```diff
@@ -80,6 +80,7 @@ class ProcessInfo : public TObject {
   bool IsWeakNC                  (void) const;
   bool IsWeakMix                 (void) const;
   bool IsDarkMatter              (void) const;
+  bool IsExternalGen             (void) const;
```
mroda88, line 83: I'll try to find a way to not mix physics and framework.

src/Framework/Interaction/ScatteringType.h
```diff
@@ -51,7 +51,8 @@ typedef enum EScatteringType {
   kScIMDAnnihilation,
   kScDarkMatterElastic = 101,
   kScDarkMatterDeepInelastic,
-  kScDarkMatterElectron
+  kScDarkMatterElectron,
+  kScExternalGen = 900
```
mroda88, line 55: probably it is better to define this negative as it's not physics

src/Tools/EvtLib/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,309 @@
...
+  for(const EvtLibParticle& part: rec->parts) if(pdg::IsLepton(part.pdg)) ++nLep;
+
+  const int firstLep = (nLep == 0) ? -1 : 2;
+  const int lastLep  = (nLep == 0) ? -1 : 1+nLep;
+
+  const int firstHad = (rec->parts.size() == nLep) ? -1 : 2+nLep;
+  const int lastHad  = (rec->parts.size() == nLep) ? -1 : 1+rec->parts.size();
+
+  // Neutrino is a parent to the lepton(s)
+  event->AddParticle(init_state.ProbePdg(),
```
mroda88, line 72: If we do this, we destroy the whole point of this exercise because we bypass the flux drivers that specifies the direction

src/Tools/EvtLib/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,309 @@
...
+    << ", pdg = " << tgt_pdgc << "]";
+
+  // Nucleus is a parent to everything else
+  event->AddParticle(tgt_pdgc,
+                     kIStInitialState,
+                     -1, -1, firstHad, lastHad,
+                     0, 0, 0, PDGLibrary::Instance()->Find(tgt_pdgc)->Mass(),
+                     0, 0, 0, 0);
+
+  // Include nuclear target in this position (2) by convention
```
mroda88, line 95: This is not always the case, it should be dropped and it certainly it is not supposed to be taken from the Summary at this stage

mroda88 (changes requested):
There a few major mistakes that makes the whole point of this library not even useful. I'll work on the fixes

src/Tools/EvtLib/EventLibraryInterface.cxx
```diff
@@ -0,0 +1,298 @@
...
+}
+
+//____________________________________________________________________________
+std::vector<TVector3> EventLibraryInterface::Basis(const TVector3& z) const
+{
+  TVector3 up(0, 1, 0);
+  if(up.Dot(z) == 0) up = TVector3(1, 0, 0);
+
+  const TVector3 x = up.Cross(z).Unit(); // Perpendicular to neutrino and up
+  const TVector3 y = x.Cross(z).Unit();  // Defines the third axis
```
mroda88, line 175: this should be -y

mroda88:
Merged manually

PR #105, opened 2020-06-05
Fix issue with electron (anti)neutrino scattering on helium

igor144:
The Smith-Monitz model (like any other RFG model) is not applicable for such nuclei as He3 and He4, so we have never tested it for these cases. We consider a reaction of a neutrino with H1, H2 and H3 as a special case, so for these nuclei, everything works. We suggest an ad-hoc solution to the problem with calculation cross-section near to the threshold.
1. Set Q2_min = Q2_max instead of exit(1) and print the corresponding message. This will set the cross-section equal zero at near-threshold energies.
2. Change the binding energy for He3 and He4 and set them equal to 7.5 MeV and 27.4 MeV respectively.


igor144:
Close, because of new pull request will be open

PR #106, opened 2020-06-09
Some small correction to BSKLN and Smith-Monitz models

igor144:
There is a problem with the calculation of electron (anti)neutrino scattering on He3 and He4.
The Smith-Monitz model (like any other RFG model) is not applicable for such nuclei as He3 and He4, so we have never tested it for these cases. We consider a reaction of a neutrino with H1, H2 and H3 as a special case, so for these nuclei, everything works. The problem arises in the near-threshold region and is related to the kinematics, which depends on binding energies for He3 and He4.
We suggest an ad-hoc solution to the problem with calculation cross-section near to the threshold.
Set Q2_min = Q2_max instead of exit(1) and print the corresponding message. This will set the cross-section equal zero at near-threshold energies.
Change the binding energy for He3 and He4 and set them equal to 7.5 MeV and 27.4 MeV respectively according to most recent experimental data.

Speed up the integration for Smith-Monitz model.

Fix two bugs with signs for BSKLN model. These bugs do not affect anything, because they eliminate each other. It is done for order.


igor144:
Will be committed again with small corrections.

PR #109, opened 2020-07-02
Fix mistype bug in LinkDef.h

igor144:


igor144:
It related to previous commit, without it there is an error 'GENIE ROOT5 on slf6 Failed at phase buildcore'

PR #115, opened 2020-08-27
NC COH gamma generator

mroda88:


mroda88:
This pull request is now closed as a number of issues were noticed. A new one will be open that contains the dedicated fixes.

PR #117, opened 2020-10-09
Free nucleon paper

jtenavidal:
Updating the xml files to be consistent with the paper results. This should be included in version 3.2


mroda88 (approved):
ok

PR #124, opened 2020-10-30
fix to gFNALExptEvGen and gT2KEvGen and replace some exit(1) calls with exception throws

nusense:
Simple changes to handle variable bin width flux histograms
change two instances of exit(1) to an exception throw 
I also added one more case to the Jacobian (if the user happens to use "to" == "from" just return 1).


mroda88 (approved):
Great!

PR #133, opened 2020-12-18
Devel 02 11a config fix

afropapp13:
Resolving issue with GXY_AB_02_11a tunes. Coping the KNO2Pythia XML block into the tune-specific CommonParam.xml files while leaving everything else the same

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-bin/
+
+mySplines/bin/
 lib/
 doxygen/
 build-env.log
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
-bin/
+
+mySplines/bin/
```
mroda88, line 2: Can you avoid this change?

mroda88 (approved):
Just perfect! thanks

PR #138, opened 2021-02-16
Devel gst

afropapp13:
Adding xsec, differential xsec & phase space choice in the gst file.

### DIFF
--- a/src/Apps/gNtpConv.cxx
+++ b/src/Apps/gNtpConv.cxx
@@ -398,11 +398,10 @@ void ConvertToGST(void)
                                  //  - ((e/h) * energy)   for pi0, gamma, e-, e+, where e/h is set to 1.3
                                  //  - (kinetic energy) for other particles
 
-  double brXSec;              // the event cross section in 1E-38cm^2
-  double brDXSec;             // is the differential cross section for the selected in 1E-38cm^2/{K^n}
-  int    brKPS;               // phase space that the xsec has been evaluated into as defined in
-                              // https://github.com/afropapp13/Generator/blob/master/src/Framework/Conventions/KinePhaseSpace.h
-
+  Double_t  brXSec;              // the event cross section in 1E-38cm^2
+  Double_t  brDXSec;             // is the differential cross section for the selected in 1E-38cm^2/{K^n}
+  UInt_t    brKPS;               // phase space that the xsec has been evaluated into
+                              
   // Open output file & create output summary tree & create the tree branches
   //
   LOG("gntpc", pNOTICE) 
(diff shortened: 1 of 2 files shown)


src/Apps/gNtpConv.cxx
```diff
@@ -398,6 +398,11 @@ void ConvertToGST(void)
                                  //  - ((e/h) * energy)   for pi0, gamma, e-, e+, where e/h is set to 1.3
                                  //  - (kinetic energy) for other particles
 
+  double brXSec;              // the event cross section in 1E-38cm^2
+  double brDXSec;             // is the differential cross section for the selected in 1E-38cm^2/{K^n}
+  int    brKPS;               // phase space that the xsec has been evaluated into as defined in
+                              // https://github.com/afropapp13/Generator/blob/master/src/Framework/Conventions/KinePhaseSpace.h
```
mroda88, line 404: can we get rid of this link?

src/Apps/gNtpConv.cxx
```diff
@@ -398,6 +398,11 @@ void ConvertToGST(void)
                                  //  - ((e/h) * energy)   for pi0, gamma, e-, e+, where e/h is set to 1.3
                                  //  - (kinetic energy) for other particles
 
+  double brXSec;              // the event cross section in 1E-38cm^2
+  double brDXSec;             // is the differential cross section for the selected in 1E-38cm^2/{K^n}
+  int    brKPS;               // phase space that the xsec has been evaluated into as defined in
```
mroda88, line 403: I would prefer this to be an `unsigned int`. So, I would prefer this to be `UInt_t`

src/Apps/gNtpConv.cxx
```diff
@@ -398,6 +398,11 @@ void ConvertToGST(void)
                                  //  - ((e/h) * energy)   for pi0, gamma, e-, e+, where e/h is set to 1.3
                                  //  - (kinetic energy) for other particles
 
+  double brXSec;              // the event cross section in 1E-38cm^2
```
mroda88, line 401: Following https://root.cern.ch/doc/master/classTTree.html I would recommend to define these variables as Double_t

src/Apps/gNtpConv.cxx
```diff
@@ -398,6 +398,11 @@ void ConvertToGST(void)
...
+  double brDXSec;             // is the differential cross section for the selected in 1E-38cm^2/{K^n}
```
mroda88, line 402: same here

src/Apps/gNtpConv.cxx
```diff
@@ -502,6 +507,9 @@ void ConvertToGST(void)
   s_tree->Branch("vtxt",         &brVtxT,	    "vtxt/D"        );
   s_tree->Branch("sumKEf",       &brSumKEf,	    "sumKEf/D"      );
   s_tree->Branch("calresp0",     &brCalResp0,	    "calresp0/D"    );
+  s_tree->Branch("XSec",         &brXSec,	    "XSec/D"    );
+  s_tree->Branch("DXSec",         &brDXSec,	    "DXSec/D"    );
+  s_tree->Branch("KPS",          &brKPS,	    "KPS/I"    );
```
mroda88, line 512: Make it an unsigned int, `"KPS/i"`

mroda88 (changes requested):
There are minor changes, but the fix seems trivial to me. 
@candreop are you ok with adding cross sections into the gst tree?

candreop:
Sounds ok @mroda88. Though I am sure that sooner or later we will see a github issue with a plot of the differential cross-section, straight out of gst, from a user who is not aware that, say, the stored values for resonance and DIS events are not comparable (without a jacobian)

mroda88 (approved):
It's good for me

PR #140, opened 2021-02-25
for COH interaction, the ConfigPool flag should be CC-PION/NC-PION, n…

afropapp13:
For COH interactions, the config flash should be CC-PION / NC-PION, not CC-Default/NC-Default
Such an issue was identified in the GTEST19_10b tune


mroda88 (approved):
good, thanks for the fix

PR #141, opened 2021-03-01
Origin/develop/lyon farm

jtenavidal:
This is the pull request to add the new information for the lyon farm queue system (GENERATOR). Will test today


src/scripts/production/batch/rmv/submit_km3net_xsec_calc_jobs.pl
```diff
@@ -56,6 +56,9 @@
 $genie_setup    = "$softw_topdir/builds/$arch/$genie_version-setup";
 $jobs_dir       = "$softw_topdir/scratch/xsec\_km3net-$production\_$cycle/";
 $freenucsplines = "$softw_topdir/data/job_inputs/xspl/gxspl-vN-UHE-$genie_version.xml";
+if ( $batch_system eq 'LyonPBS' ) {
+    $queue = "P_gdrnu_genie" ;
```
mroda88, line 60: No, this makes it impossible for the user to configure this variable. The right thing to do is: 
1) you first set the batch system variable
2) based on the batch system you defined a `$default_queue` 
3) you assign the default value to `$queue` with the caveat `unless defined $queue`.

mroda88 (changes requested):
Hi Julia, no the logic implemented is faulted. Check the message. 
Also, I think we can forget about the modules we are not using. Most of them are not supported. I think from the generator we only care about the vN and vA splines.

mroda88 (approved):
perfect

PR #142, opened 2021-03-09
Changing event generation list naming to include new physics processes

afropapp13:
Changing event generation list naming to include new physics processes by default (like MEC & DFR)
Simplifying the naming scheme, that is "CC", "NC" and "EM" instead of "CCinclMEC" and "EM+MEC" [there was no inclusive process list for NC]


mroda88 (changes requested):
We are definitely in the right direction. But the CC is missing the `genie::EventGenerator/QEL-CC-LAMBDA`.

mroda88 (approved):
Sound!

PR #148, opened 2021-03-30
Fix compilation problem for the DarkSectorDecayer

sjgardiner:
Switch to using a vector rather than a C-style array to store the particle masses for the DarkSectorDecayer. This resolves a compilation error for GCC 4.8.5 and manages memory in a safer way.


mroda88 (approved):
Perfect

Issue #149, opened 2021-03-30
genie exits immediately when installation path contains some characters


davehadley:
If GENIE is installed in a path that happens to contain the characters `-l` then the  `$GENIE/bin/genie` application exits immediately with the error message:

```
Failed to load GENIE libraries
```

The problem is caused by code that parses libraries returned by `genie-config --libs` thinks that the directory is a library to be loaded. When it fails to load that, it exits.

The problem is caused by the line in: [loadlibs.C](https://github.com/GENIE-MC/Generator/blob/master/src/scripts/gcint/loadlibs.C).

```
TPRegexp re("-l([\\d\\w]*)");
```

This can be made slightly more robust by modifying the reg-ex:

```
TPRegexp re("^-l([\\d\\w]*)");
```

but I guess that this will still fail in the unlikely case that the GENIE path begins with the characters `-l`.

Issue #145, opened 2021-03-25
Hidden ROOT 5 dependency


mroda88:
We might have dropped the ROOT 5 dependency, but our Makefiles are still supporting it.

Specifically we have https://github.com/GENIE-MC/Generator/blob/c372775dda9589b4a3a43826d74bc88e331933be/src/make/Make.std-package-targets#L54 and following lines. 

Can we clean this file. 
Also we need to update the make file of comparisons as well. I'll open a dedicated issue there

Issue #116, opened 2020-10-04
GENIE deletes operating system


arguelles:
When doing make cleandist, GENIE decides to, to first order, delete the OS.
![Captura de Pantalla 2020-10-04 a la(s) 10 32 42 a  m](https://user-images.githubusercontent.com/7256191/95018708-58530f00-062f-11eb-83e6-fdbbae02eafc.png)

tamasgal:
This issue probably deserves a bit more attention 😉

tamasgal:
This is the line which will operate on the root filesystem level when some variables are emtpy: https://github.com/GENIE-MC/Generator/blob/0b80f5ca6b315f5f7138c0fdf60ce40a4997697c/src/make/Make.std-package-targets#L172

PR #153, opened 2021-04-15
resolve a number of small issues

nusense:
This pull request should resolve issues
#149 loadlibs.C failing when "-l" installation path;  match "-l" at beginning of token
#116 don't perform "distclean" on /usr or /usr/local
#145 remove leftover ROOT5 cruft
it also remove -ffriend-injection flag (deprecated) when using gcc >= 8.0.0
and loadlibs.C also handles -L/path/dir from genie-config --libs (and a small potential memory leak)
for INCL++ look for non-mt boost library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,6 +344,14 @@ endif
 make-install-dirs: FORCE
 	@echo " "
 	@echo "** Creating directory structure for GENIE installation..."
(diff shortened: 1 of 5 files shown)


mroda88 (approved):
Great!

PR #160, opened 2021-05-17
Add Valencia-style binding energy treatment as an option for QE event generation

sjgardiner:
The effective energy transfer defined in Eq. (43) of the original Valencia model paper (https://arxiv.org/abs/nucl-th/0408005) may be recast in terms of a particular choice of an initial struck nucleon total energy (possibly off the mass shell). For EMQE and NCEL, the treatment leaves the nucleon on-shell. For CCQE, there is a modest correction to the on-shell value.

This pull request adds the Valencia binding energy "recipe" as an option to use when generating CCQE events. It lives alongside the others harmoniously, including the default "UseNuclearModel" option.


mroda88 (approved):
This looks good to me.
Let's merge according to what we discussed at the forum

PR #162, opened 2021-05-27
Qvalue shift new

jtenavidal:
New implementation of the Qvalue shift

### DIFF
--- a/config/QvalueShifter.xml
+++ b/config/QvalueShifter.xml
@@ -9,6 +9,7 @@ Name                   Type     Optional   Comment                              
 ......................................................................................................................
 QvalueShiftDefault     double   No         Default Qvalue Shift for all targets          0.
 QvalueShift@Pdg=PDG    double   Yes        Add possible Relative Qvalue shifts given a PDG code
+                                           Values can be positive or negative
 -->
 
   <param_set name="Default"> 
@@ -16,11 +17,11 @@ QvalueShift@Pdg=PDG    double   Yes        Add possible Relative Qvalue shifts g
   </param_set>
 
   <param_set name="1p1h"> 
-    <param type="double" name="QvalueShift@Pdg=1000060120">  1.2 </param> <!-- C12    -->
+    <param type="double" name="QvalueShift@Pdg=1000060120">  1.2 </param> <!-- C12 - this is the recomended max variation -->
   </param_set>
 
   <param_set name="2p2h"> 
-    <param type="double" name="QvalueShift@Pdg=1000060120">  2.4 </param> <!-- C12    -->
+    <param type="double" name="QvalueShift@Pdg=1000060120">  2.4 </param> <!-- C12 - this is the recomended max variation -->
   </param_set>
 
 </alg_conf>
--- a/config/QvalueShifter.xml
+++ b/config/QvalueShifter.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<alg_conf>
+<!--
+Configuration for the QvalueShifter
+Configurable Parameters:
+......................................................................................................................
+Name                   Type     Optional   Comment                                       Default
+......................................................................................................................
+QvalueShiftDefault     double   No         Default Qvalue Shift for all targets          0.
+QvalueShift@Pdg=PDG    double   Yes        Add possible Relative Qvalue shifts given a PDG code
+-->
+
+  <param_set name="Default"> 
+    <param type="double"  name="QvalueShiftDefault"> 0. </param>   
+  </param_set>
+
+  <param_set name="1p1h"> 
+    <param type="double" name="QvalueShift@Pdg=1000060120">  0. </param> <!-- C12    -->
+    <param type="double" name="QvalueShift@Pdg=1000080160">  0. </param> <!-- O16    -->
+    <param type="double" name="QvalueShift@Pdg=1000180400">  0. </param> <!-- Ar40   -->
+    <param type="double" name="QvalueShift@Pdg=1000260560">  0. </param> <!-- Fe56   -->
+  </param_set>
+
+  <param_set name="2p2h"> 
+    <param type="double" name="QvalueShift@Pdg=1000060120">  0. </param> <!-- C12    -->
+    <param type="double" name="QvalueShift@Pdg=1000080160">  0. </param> <!-- O16    -->
+    <param type="double" name="QvalueShift@Pdg=1000180400">  0. </param> <!-- Ar40   -->
+    <param type="double" name="QvalueShift@Pdg=1000260560">  0. </param> <!-- Fe56   -->
+  </param_set>
+
+</alg_conf>
--- a/config/QvalueShifter.xml
+++ b/config/QvalueShifter.xml
@@ -16,17 +16,11 @@ QvalueShift@Pdg=PDG    double   Yes        Add possible Relative Qvalue shifts g
   </param_set>
 
   <param_set name="1p1h"> 
-    <param type="double" name="QvalueShift@Pdg=1000060120">  0. </param> <!-- C12    -->
-    <param type="double" name="QvalueShift@Pdg=1000080160">  0. </param> <!-- O16    -->
-    <param type="double" name="QvalueShift@Pdg=1000180400">  0. </param> <!-- Ar40   -->
-    <param type="double" name="QvalueShift@Pdg=1000260560">  0. </param> <!-- Fe56   -->
+    <param type="double" name="QvalueShift@Pdg=1000060120">  1.2 </param> <!-- C12    -->
   </param_set>
 
   <param_set name="2p2h"> 
-    <param type="double" name="QvalueShift@Pdg=1000060120">  0. </param> <!-- C12    -->
-    <param type="double" name="QvalueShift@Pdg=1000080160">  0. </param> <!-- O16    -->
-    <param type="double" name="QvalueShift@Pdg=1000180400">  0. </param> <!-- Ar40   -->
-    <param type="double" name="QvalueShift@Pdg=1000260560">  0. </param> <!-- Fe56   -->
+    <param type="double" name="QvalueShift@Pdg=1000060120">  2.4 </param> <!-- C12    -->
   </param_set>
 
 </alg_conf>
--- a/src/Physics/Common/QvalueShifter.cxx
+++ b/src/Physics/Common/QvalueShifter.cxx
@@ -105,7 +105,7 @@ void QvalueShifter::LoadConfig(void)
   }
 
   if( ! good_config ) {
-    LOG("QvalueShifter", pERROR) << "QvalueShifter Configuration has failed.";
+    LOG("QvalueShifter", pERROR) << "Configuration has failed.";
     exit(78) ;
   }
 
--- a/src/Physics/Common/QvalueShifter.cxx
+++ b/src/Physics/Common/QvalueShifter.cxx
@@ -74,6 +74,7 @@ void QvalueShifter::Configure(string config)
 
 void QvalueShifter::LoadConfig(void)
 {
+  bool good_config = true ;
   // Store default value
   GetParam( "QvalueShiftDefault", fRelShiftDefault, 0. ) ;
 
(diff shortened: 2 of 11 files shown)


config/QvalueShifter.xml
```diff
@@ -0,0 +1,27 @@
...
+QvalueShift@Pdg=PDG    double   Yes        Add possible Relative Qvalue shifts given a PDG code
+                                           Values can be positive or negative
+-->
+
+  <param_set name="Default"> 
+    <param type="double"  name="QvalueShiftDefault"> 0. </param>   
+  </param_set>
+
+  <param_set name="1p1h"> 
+    <param type="double" name="QvalueShift@Pdg=1000060120">  1.2 </param> <!-- C12 - this is the recomended max variation -->
```
mroda88, line 20: Just a cross check. The maximum shift is 120 % ? 
what about the minimum?

src/Physics/Multinucleon/XSection/NievesSimoVacasMECPXSec2016.cxx
```diff
@@ -339,4 +342,7 @@ void NievesSimoVacasMECPXSec2016::LoadConfig(void)
         dynamic_cast<const XSecIntegratorI *> (
           this->SubAlg("NumericalIntegrationAlg"));
         assert(fXSecIntegrator);
+
+	// Read optional QvalueShifter:
+	fQvalueShifter = dynamic_cast<const QvalueShifter *> ( this->SubAlg("QvalueShifterAlg") );
```
mroda88, line 347: This is a bit too simple.
1) initialise the pointer to nullptr
2) check if the Registry contains a shifter algo
3) In case of (2) set the pointer and check that the dynamic cast went through, if not report an error because the desired configuration cannot be satistifed.

src/Physics/Multinucleon/XSection/SuSAv2MECPXSec.cxx
```diff
@@ -379,4 +382,9 @@ void SuSAv2MECPXSec::LoadConfig(void)
   this->GetParam( "RFG-NucRemovalE@Pdg=1000501190", fEbSn );
   this->GetParam( "RFG-NucRemovalE@Pdg=1000791970", fEbAu );
   this->GetParam( "RFG-NucRemovalE@Pdg=1000822080", fEbPb );
+
+
+  // Read optional QvalueShifter:
+  fQvalueShifter = dynamic_cast<const QvalueShifter *> ( this->SubAlg("QvalueShifterAlg") );
```
mroda88, line 388: Same as the comment on Nieves

src/Physics/QuasiElastic/XSection/SuSAv2QELPXSec.cxx
```diff
@@ -308,4 +311,7 @@ void SuSAv2QELPXSec::LoadConfig(void)
   this->GetParam( "RFG-NucRemovalE@Pdg=1000791970", fEbAu );
   this->GetParam( "RFG-NucRemovalE@Pdg=1000822080", fEbPb );
 
+  // Read optional QvalueShifter:
+  fQvalueShifter = dynamic_cast<const QvalueShifter *> ( this->SubAlg("QvalueShifterAlg") );
```
mroda88, line 315: Same comment as per Nieves

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+    this->LoadConfig();
+}
+//_________________________________________________________________________
+
+  double QvalueShifter::Shift( const Target & target ) const {
+    // Get Target pdg
+    int pdg_target = target.Pdg() ;
+
+    // Find Pdg in map:
+    if ( fRelShift.find(pdg_target) != fRelShift.end() ) {
```
mroda88, line 54: Write it more efficiently.
You loop over the map twice with your code: first to check if the pdg is there, second to return the value.
Just save the returned value in the first place:
```C++
const auto it = fRelShift.find(pdg_target) ;
if ( it != fRelShift.end() ) {
  return it -> second ;
}
```

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+  } 
+
+//_________________________________________________________________________
+
+  double QvalueShifter::Shift( const Interaction & interaction ) const {
+    // This function allows the flexibility to add shift as a function of 
+    // the interaction type.
+    // Right now, we just call the default :
+
+    return Shift( * interaction.InitState().TgtPtr() ) ;
```
mroda88, line 70: Avoid one useless dereference: 
```C++
  return Shift( interaction.InitState().Tgt() ) ;
```

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+    
+  } 
+//_________________________________________________________________________
+
+void QvalueShifter::LoadConfig(void)
+{
+  // Store default value
+  GetParam( "QvalueShiftDefault", fRelShiftDefault, 0. ) ;
```
mroda88, line 78: No, if you set the default, you have to call GetParamDef. This is wrong becuase the 0 is interpreted as a boolean.

mroda88: Also the `xml` file says it is not optional, so just 
```C++
 GetParam( "QvalueShiftDefault", fRelShiftDefault) ;
```

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+
+  // Get possible entries to pdg - shift map 
+  RgKeyList kpdg_list = GetConfig().FindKeys("QvalueShift@Pdg=") ;
+  RgKeyList::const_iterator kiter = kpdg_list.begin();
```
mroda88, line 82: Put this inside the loop

mroda88: You might want to consider using `auto`

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+  for( ; kiter != kpdg_list.end(); ++kiter ) {
+    RgKey key = *kiter ; 
+    double rshift = GetConfig().GetDouble(key);
+    vector<string> kv = genie::utils::str::Split(key,"=");
+    assert(kv.size()==2);
+    int pdg_target = atoi(kv[1].c_str());
```
mroda88, line 88: I forbid you to use `atoi` if you can avoid it. It's an old heritage from dark ages. Please avoid it and use something new.
Use either `stoi` https://www.cplusplus.com/reference/string/stoi/ or a `stringstream` to do the conversion.

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+    // the interaction type.
+    // Right now, we just call the default :
+
+    return Shift( * interaction.InitState().TgtPtr() ) ;
+    
+  } 
+//_________________________________________________________________________
+
+void QvalueShifter::LoadConfig(void)
+{
```
mroda88, line 76: Before everything else do a clear of the map in case there were previous configurations

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+  // Store default value
+  GetParam( "QvalueShiftDefault", fRelShiftDefault, 0. ) ;
+
+  // Get possible entries to pdg - shift map 
+  RgKeyList kpdg_list = GetConfig().FindKeys("QvalueShift@Pdg=") ;
+  RgKeyList::const_iterator kiter = kpdg_list.begin();
+  for( ; kiter != kpdg_list.end(); ++kiter ) {
+    RgKey key = *kiter ; 
+    double rshift = GetConfig().GetDouble(key);
```
mroda88, line 85: About the logic, why don't you simply
```C++
  int pdg_target = stoi( kv[1] );
  GetParam( key, fRelShift[pdg_target] )
```
it seems much simpler that all that fancy pair contraption.

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,93 @@
...
+void QvalueShifter::LoadConfig(void)
+{
+  // Store default value
+  GetParam( "QvalueShiftDefault", fRelShiftDefault, 0. ) ;
+
+  // Get possible entries to pdg - shift map 
+  RgKeyList kpdg_list = GetConfig().FindKeys("QvalueShift@Pdg=") ;
+  RgKeyList::const_iterator kiter = kpdg_list.begin();
+  for( ; kiter != kpdg_list.end(); ++kiter ) {
+    RgKey key = *kiter ;
```
mroda88, line 84: This can be a reference
```C++
  const RgKey & key =  *kiter ;
```
no need to create pointless copies

src/Physics/Common/QvalueShifter.cxx
mroda88: I would also check if the PDG is a valid isotope. If not call for errors.

mroda88 (changes requested):
Looks promising, but it's a bit doggy in some places.

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,99 @@
...
+
+  // Get possible entries to pdg - shift map 
+  auto kpdg_list = GetConfig().FindKeys("QvalueShift@Pdg=") ;
+
+  for( auto kiter = kpdg_list.begin(); kiter != kpdg_list.end(); ++kiter ) {
+    const RgKey & key = *kiter ;
+    vector<string> kv = genie::utils::str::Split(key,"=");
+    assert(kv.size()==2);
+    int pdg_target = stoi( kv[1] );
+    if( ! PDGLibrary::Instance()->Find(pdg_target) ) {
```
mroda88, line 91: Not only. You also have to check if it is a nucleus PDG.

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,99 @@
...
+      LOG("NievesSimoVacasMECPXSec2016", pERROR) << "The Pdg code associated to the QvalueShift is not valid : " << pdg_target ; 
+      continue ; 
+    }
+    GetParam( key, fRelShift[pdg_target] ) ;
+  }
+}
```
mroda88, line 97: If there is an error, you should call for exit() and use the appropriate error code, as we do in other modesl.

src/Physics/Common/QvalueShifter.cxx
```diff
@@ -0,0 +1,99 @@
...
+  // Get possible entries to pdg - shift map 
+  auto kpdg_list = GetConfig().FindKeys("QvalueShift@Pdg=") ;
+
+  for( auto kiter = kpdg_list.begin(); kiter != kpdg_list.end(); ++kiter ) {
+    const RgKey & key = *kiter ;
+    vector<string> kv = genie::utils::str::Split(key,"=");
+    assert(kv.size()==2);
+    int pdg_target = stoi( kv[1] );
+    if( ! PDGLibrary::Instance()->Find(pdg_target) ) {
+      LOG("NievesSimoVacasMECPXSec2016", pERROR) << "The Pdg code associated to the QvalueShift is not valid : " << pdg_target ;
```
mroda88, line 92: It's not NievesSimoVavas, it's "QValueShifter"

src/Physics/Multinucleon/XSection/NievesSimoVacasMECPXSec2016.cxx
```diff
@@ -328,15 +331,36 @@ void NievesSimoVacasMECPXSec2016::Configure(string config)
...
+    LOG("NievesSimoVacasMECPXSec2016", pERROR) << "The required NumericalIntegrationAlg does not exist : " << SubAlg("NumericalIntegrationAlg");
+  }
+  
+  // Read optional QvalueShifter:
+  fQvalueShifter = nullptr; 
+  if( GetConfig().Exists("QvalueShifterAlg") ) {
+    fQvalueShifter = dynamic_cast<const QvalueShifter *> ( this->SubAlg("QvalueShifterAlg") );
+    if( !fQvalueShifter ) {
+      good_config = false ; 
+      LOG("NievesSimoVacasMECPXSec2016", pERROR) << "The required QvalueShifterAlg does not exist : " << SubAlg("QvalueShifterAlg") ;
```
mroda88, line 357: The right thing to pass here is the AlgoID associated to the key ""QvalueShifterAlg" not the algo itself.

src/Physics/Multinucleon/XSection/SuSAv2MECPXSec.cxx
```diff
@@ -379,4 +388,21 @@ void SuSAv2MECPXSec::LoadConfig(void)
...
   this->GetParam( "RFG-NucRemovalE@Pdg=1000822080", fEbPb );
+
+
+  // Read optional QvalueShifter:
+  fQvalueShifter = nullptr; 
+  if( GetConfig().Exists("QvalueShifterAlg") ) {
+    fQvalueShifter = dynamic_cast<const QvalueShifter *> ( this->SubAlg("QvalueShifterAlg") );
+    if( !fQvalueShifter ) {
+      good_config = false ; 
+      LOG("SuSAv2MECPXSec", pERROR) << "The required QvalueShifterAlg does not exist : " << SubAlg("QvalueShifterAlg") ;
```
mroda88, line 399: Same as per Nieves

mroda88 (changes requested):
You didn't fix all the things julia and there were other mistakes.

mroda88 (approved):
Ok, let's merge it.

PR #166, opened 2021-06-08
Adding Hadronization paper CMC

jtenavidal:
In this pull request, the different CMC containing the results from the hadronization paper are handled. So far I added the configurations for the G18_02a only.


mroda88 (approved):
ok

PR #170, opened 2021-08-02
Develop/fsi reweight

jtenavidal:
This class loops over the event record to find particles that were affected by FSI. Depending on their fate, we apply a weight to the event. This option can be configured using the corresponding xml files. The code is generic given that the HadronTransporter model sets the HN Fates.

### DIFF
--- a/config/CascadeReweight.xml
+++ b/config/CascadeReweight.xml
@@ -5,17 +5,47 @@
 <!--
 Configuration sets for the CascadeReweight EventRecordVisitorI
 
+..............................................................................................................
 Configurable Parameters:
-............................................................................................
-Name         Type     Optional   Comment                     Default
-............................................................................................
-
-
+..............................................................................................................
+Name                                 Type     Optional   Comment                                  Default
+..............................................................................................................
+CascadeReweight-Default-Weight       double   No         Default weight associated to a FSI fate
+CascadeReweight-Weight-Undefined     double   Yes        Weight associated to the undefined fate
+CascadeReweight-Weight-NoInteraction double   Yes        Weight associated to the nointeraction fate
+CascadeReweight-Weight-CEx           double   Yes        Weight associated to the CEx fate
+CascadeReweight-Weight-Elastic       double   Yes        Weight associated to the Elastic fate
+CascadeReweight-Weight-Inelastic     double   Yes        Weight associated to the Inelastic fate
+CascadeReweight-Weight-Abs           double   Yes        Weight associated to the Abs fate
+CascadeReweight-Weight-Cmp           double   Yes        Weight associated to the Cmp fate
+..............................................................................................................
 -->
 
   <param_set name="Default"> 
-
+    <param type="double"  name="CascadeReweight-Default-Weight"> 1. </param>   
   </param_set>
+  
+  <param_set name="PionReweight">
+    <param type="double" name="CascadeReweight-Weight-Undefined@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-NoInteraction@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-CEx@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-Abs@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-Elastic@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-Inelastic@Pdg=211"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-Cmp@Pdg=211"> 1. </param>
+  </param_set>
+
+  <param_set name="ProtonReweight">
+    <param type="double" name="CascadeReweight-Weight-Undefined@Pdg=2212"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-NoInteraction@Pdg=2212"> 1. </param>
+    <param type="double" name="CascadeReweight-Weight-CEx@Pdg=2212"> 1. </param>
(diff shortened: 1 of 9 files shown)


jtenavidal:
I tested it locally and it follows the expected behavior. I will now test it with GENIE Comparisons. Focus on Minerva data, pi+ absorbtion and pion inel events. Will add the results here when available.

src/Physics/HadronTransport/CascadeReweight.cxx
```diff
@@ -0,0 +1,180 @@
...
+  Algorithm::Configure(param_set);
+  this->LoadConfig();
+}
+//___________________________________________________________________________
+void CascadeReweight::LoadConfig(void)
+{
+  bool good_config = true ; 
+
+  // Get default weight 
+  GetParamDef( "CascadeReweight-Default-Weight", fDefaultWeight, 1. ) ;
```
mroda88, line 118: this needs to go

src/Physics/HadronTransport/INukeHadroFates2018.h
```diff
@@ -116,6 +116,15 @@ class INukeHadroFates {
      return "undefined"; 
   }
   //__________________________________________________________________________
+
+  static std::map<INukeFateHN_t,string> GetEINukeFateKeysMap( void ) {
```
mroda88, line 120: Because this map is inconsistent with the rest, we should move it to the CascateReweight header.

mroda88: since the map itself is static, you can probably return a reference in this case.

src/Physics/HadronTransport/INukeHadroFates2018.h
```diff
@@ -116,6 +116,15 @@ class INukeHadroFates {
...
+    static std::map<INukeFateHN_t,string> map_keys { {kIHNFtUndefined,"Undefined"}, {kIHNFtNoInteraction,"NoInteraction"},
```
mroda88, line 121: the map has to `const` and so the returned type

src/Physics/HadronTransport/HadronTransporter.cxx
```diff
@@ -154,7 +165,26 @@ void HadronTransporter::LoadConfig(void)
...
+     if( GetConfig().Exists("CascadeReweightAlg") ) {
+       fCascadeReweight = dynamic_cast<const CascadeReweight *> ( this->SubAlg("CascadeReweightAlg") );
+       if( !fCascadeReweight ) {
+	 good_config = false ; 
+	 LOG("HadronTransporter", pERROR) << "The required CascadeReweightAlg does not exist. AlgID is : " << SubAlg("CascadeReweightAlg")->Id() ;
+       }
+     }
+
+     if( ! good_config ) {
+       LOG("HadronTransporter", pERROR) << "Configuration has failed.";
```
mroda88, line 185: This is a `pFATAL`

src/Physics/HadronTransport/CascadeReweight.cxx
```diff
@@ -0,0 +1,180 @@
...
+  
+  // Clean maps
+  fDefaultMap.clear(); 
+  fFateWeightsMap.clear();
+
+  // Create vector with list of possible keys (follows the order of the fates enumeration)
+  std::map<INukeFateHN_t,string> EINukeFate_map_keys = INukeHadroFates::GetEINukeFateKeysMap() ;
```
mroda88, line 125: The new map now will be in the CascadeReweight namespace

src/Physics/HadronTransport/CascadeReweight.cxx
```diff
@@ -0,0 +1,180 @@
...
+	LOG("CascadeReweight", pERROR) << "The weight assigned to " << to_find_def << " is not positive" ; 
+	good_config = false ; 
+	continue ; 
+      }
+      fDefaultMap[it_keys->first] = weight ; 
+    }
+
+    // Find Pdg specifications
+    std::string to_find_pdg = "CascadeReweight-Weight-"+(it_keys->second)+"@Pdg=" ;
+    std::map<int,double> WeightMap ; // define map that stores <pdg, weight>
```
mroda88, line 147: This variable should be defined inside the loop

src/Physics/HadronTransport/CascadeReweight.cxx
```diff
@@ -0,0 +1,180 @@
...
+      if( weight < 0 ) {
+	LOG("CascadeReweight", pERROR) << "The weight assigned to " << to_find_pdg << " is not positive" ; 
+	good_config = false ; 
+	continue ; 
+      }
+      // Add pdg and weight in map
+      WeightMap.insert( std::pair<int,double>( pdg_target, weight ) ) ;
+    }
+    // store information in class member
+    fFateWeightsMap[it_keys->first] = WeightMap ;
```
mroda88, line 171:
```c++
 fFateWeightsMap[it_keys->first] = std::move(WeightMap) ; 
```

mroda88 (changes requested):
glad it works. Some logic and code fixing necessary.

config/CascadeReweight.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<alg_conf>
+
+<!--
+Configuration sets for the CascadeReweight EventRecordVisitorI
```
mroda88, line 6: here add a line saying that clearly this configuration file is done for tuning purposes, and the present configurations are here only as an example

mroda88 (changes requested):
Just small fixes and then we are good to go

mroda88 (approved):
It will probably not be used anytime soon, but the interface is very clean and can be used as a base for future works.

PR #175, opened 2021-09-14
Develop/h a knobs

jtenavidal:
Adding parameters to tune pion absorbtion for charged pions and neutral pions.


mroda88 (approved):
Code as expected from discussion with Steve Dytman.

PR #179, opened 2021-10-04
Fixing default behaviour

jtenavidal:
The weight for events with W<0 is left to the default weight of 1. Because the default weight is configurable (fDefaultWeight), I decided to change this to 1 to avoid confusion.


mroda88 (approved):
ok

PR #180, opened 2021-10-05
Update SlowRsclCharmDISPXSecLO.h

nusense:
fix typo in forward declaration of XSecIntegratorI (was missing trailing "I").  Caused difficulties in dictionary creation if headers in directory were used in reverse alphabetical order.


mroda88 (approved):
It looks good to me and I also checked that the setting of that variable in LoadConfig is casting to the right interface.

PR #181, opened 2021-10-14
adding different parameters for neutral and charged pion mean free path

jtenavidal:
Adding a MFP parameter for Charged and Neutral pions. This adds coherence between the Frac Abs charged pions parameter and the MFP.


jtenavidal:
Runing a test

mroda88 (approved):
All good

PR #184, opened 2021-10-19
new App for dumping configuration in (semi)human readable form

nusense:
print out algorithm configurations and Common* configs in Registry format
list which file is being used for Common* configs


mroda88:
But the app I think it's working very well!

mroda88:
Fantastic!

PR #185, opened 2021-10-25
force TGeoManager to use kRootUnits (cm)

nusense:
ROOT versions 6.16 through 6.24 would leave the in memory representation of a GDML file in mm rather than the long standing ROOT standard of cm.  Force the use of kRootUnits before TGeoManager::Import().


mroda88 (approved):
ok

PR #186, opened 2021-10-25
Add forward declarations and using statements so that header order in dictionary creation doesn't matter

nusense:
Future versions of ROOT dictionary code will no longer include the line `namespace std {} using namespace std;`
Additionally the header order from a glob might be different


mroda88 (approved):
great!

PR #189, opened 2021-11-04
Origin/develop/mec scale vs w edges

jtenavidal:
In order to implement the new parameterization of the MECScaleVsW that we suggested in the last meeting, we need to be able to set the weight for both edges. In the previous implementation, this was the same parameter, defined by the default.

### DIFF
--- a/src/Physics/Multinucleon/XSection/MECScaleVsW.cxx
+++ b/src/Physics/Multinucleon/XSection/MECScaleVsW.cxx
@@ -88,7 +88,7 @@ MECScaleVsW::weight_type_map MECScaleVsW::GetMapWithLimits( const double Q0, con
   double max_weight = fDefaultWeight ; 
 
   if( fLowLimitWeight ) min_weight = *fLowLimitWeight ; 
-  if( fLowLimitWeight ) max_weight = *fUpperLimitWeight ; 
+  if( fUpperLimitWeight ) max_weight = *fUpperLimitWeight ; 
 
   w_map.insert( weight_type_pair( W_max, max_weight ) ) ;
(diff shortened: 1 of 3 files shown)


jtenavidal:
Code tested. Works fine.

src/Physics/Multinucleon/XSection/MECScaleVsW.cxx
```diff
@@ -147,6 +147,14 @@ void MECScaleVsW::LoadConfig(void)
     LOG("MECScaleVsW", pERROR) << "Lower limit for Q3 size: " << limit_Q3.size() ;
   }
 
+  if( GetConfig().Exists("MECScleVsW-LowerLimit-Weight") ) {
```
mroda88, line 150: Instead of this very verbose block, 
please do this
```C++
GetParamDef("MECScleVsW-LowerLimit-Weight", fLowLimitWeight, fDefaultWeight ) ; 
```

src/Physics/Multinucleon/XSection/MECScaleVsW.cxx
```diff
@@ -147,6 +147,14 @@ void MECScaleVsW::LoadConfig(void)
...
+    GetParam("MECScleVsW-LowerLimit-Weight", fLowLimitWeight ) ; 
+  } else { fLowLimitWeight = fDefaultWeight; } 
+
+  if( GetConfig().Exists("MECScleVsW-UpperLimit-Weight") ) {
```
mroda88, line 154: Same as the previous block

mroda88 (changes requested):
Please do those small changes

mroda88 (approved):
Great

PR #193, opened 2022-01-18
Update print syntax

nusense:
Make printout of syntax more consistent across apps and within the app's PrintSyntax() and it's header.  Offload some work to Framework/Utils/RunOpt to describe what it parses from the command line


mroda88 (approved):
LGTM

PR #195, opened 2022-02-04
Configurable variable bin width multiplication

nusense:
make whether variable bin flux histograms are multiplied by the bin width
  configurable from the command line
by default they are, but use of WIDTH or NOWIDTH forces the issue
Also, notify when multiplying


mroda88 (approved):
LGTM

PR #161, opened 2021-05-19
Add GGHR model to HEDIS module.

pochoarus:

PR #197, opened 2022-02-18
HEDIS updates and proper configurations

mroda88:


mroda88:
This PR is replacing https://github.com/GENIE-MC/Generator/pull/161 to avoid misconfiguration in non-HE tunes.

mroda88:
We discussed it with Alfonso.

PR #198, opened 2022-02-18
Feature/cpr2022

candreop:
Update author list in GENIE banner and bump up copyright to 2022

PR #200, opened 2022-02-21
Feature/copyright2022

mroda88:


mroda88:
This replaces #198 so that others can commit if necessary. 
I did some test and started creating different splines from many different tunes, although I didn't try them all, and all looks working for me.

PR #203, opened 2022-03-02
Adds linuxarm64 arch to make.include

luketpickering:


candreop:
@mroda88, please consider whether to patch v3.2.0 in addition to master. Since the v3.2.0 branch is not advertised yet, might be possible to include without bumping up the revision number?

mroda88:
Sure, Steven gardiner told me that he needs time to tag reweight, so we can definitely do it. @nusense can you review and merge if you like the PR?

nusense (approved):

PR #219, opened 2022-08-04
add #include RunOpt.h because otherwise treated as an incomplete class

nusense:
I'm not sure what changed, but the use of "genie::RunOpt::Instance()" was failing with v3_02_00 and root v6_22_08 because it was seen an an incomplete class.  By pulling in the header the genie-as-a-super-root works again cleanly

### DIFF
--- a/src/scripts/gcint/genie_setup.C
+++ b/src/scripts/gcint/genie_setup.C
@@ -32,6 +32,7 @@ void genie_setup()
 
     change_prompt();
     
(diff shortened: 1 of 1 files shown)


mroda88 (approved):
Looks ok to me!

mroda88:
Do we need to re-tag v3.2?

nusense:
[Sigh] I hate doing so.  I'm not sure what to do ...

PR #225, opened 2022-09-28
removing safety factor for Wcut

jtenavidal:
We noticed some double counting in electron scattering inclusive plots. This was due to a correction factor applied to fWcut. This correction  isn't needed as we are already calculating the cross sections above Wcut before merging them together. We can safely remove this additional correction factor. 
This pull request was discussed in the latest GENIE Tuning meeting (https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=320)


mroda88 (approved):
LGTM

PR #228, opened 2022-10-06
Added a QPix specific gtnpc format

delofson0211:
Added a new output format to gNtpConv.cxx specific to QPix collaboration needs.

### DIFF
--- a/src/Apps/gNtpConv.cxx
+++ b/src/Apps/gNtpConv.cxx
@@ -32,6 +32,8 @@
               >>
                * `gst': 
                     The 'definite' GENIE summary tree format (gst).
+		   * `qpix':
+		   * 		 This is for Q-Pix collaboration work at UTA
    	       * `gxml': 
                      GENIE XML event format 
    	       * `ghep_mock_data': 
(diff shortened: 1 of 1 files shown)


candreop:
Thank you David. To include this we would have to generalise some of the names (eg remove labels
such as UTAQPix). Perhaps something like "reduced_gst" might be more appropriate. But looking at what is actually stored, it seems to me that this format is very similar to the rootracker one (not the t2k or numi specific versions of the rootracker that contain a lot of flux variables, but the basic rootracker format). The rootracker format (which is some kind of fusion of the STDHEP and tracker formats used by many older experiments), like the UTAQPix one, contains only an array of particles. Rootracker also contains a few very basic variables (like the event weight, vertex position, or interaction code) needed at almost any level of analysis. You don't have the latter info, but I suspect you will need it before long. Rather that having a new format that slowly converges towards the rootracker one, is it possible that you can simply use the rootracker?

candreop:
Perhaps @nusense who looks after many of the experimental interfaces can weight in with his opinion

nusense:
The only experiment that I work with and know of that uses output of `gntpc` is DUNE NearDet, where they use `rootracker` for their input to edep-sim.  I'm not _personally_ familiar with use of anything other than the native `.ghep.root` format, so I can't really speak to this issue much, though having recently looked a bit at `rootracker` I think it's pretty barebones as it is.

delofson0211:
Ok. It sounds like rootracker should have the functionality we need. Let me look into this more in depth and see if that is adequate. We can table this discussion until I am able to confirm whether or not rootracker is sufficient.

PR #236, opened 2022-11-09
Develop/emres form factors

jtenavidal:
This pull request is based on the discussion by Afroditi at the GENIE Forum: 
https://genie-docdb.pp.rl.ac.uk/DocDB/0003/000334/005/Papadopoulou_GENIE_Core_Oct_10_2022.pdf

The main idea is that as it is now, the Form Factors used in EM mode come from a neutrino prescription. This might be problematic, so I use the alternative dipole form factors instead.

I noticed an additional potential issue - for EM the correction factor for the form factors should be different according to: 
https://genie-docdb.pp.rl.ac.uk/DocDB/0003/000334/005/tn260.pdf

This is now implemented

### DIFF
--- a/config/BergerSehgalRESPXSec2014.xml
+++ b/config/BergerSehgalRESPXSec2014.xml
@@ -74,12 +74,18 @@ XSec-Integrator             alg     No
    
   </param_set> 
 
- <param_set name="EM-WithPauliBlock"> 
+  <param_set name="EM-WithPauliBlock"> 
      <param type="alg"  name="XSec-Integrator">  genie::ReinSehgalRESXSec/WithPauliBlock    </param>
+     <!-- Use dipole form factors for electrons Ref.[F.Ravndal, Phys.Rev. D 4, 1466 (1971)] --> 
+     <param type="bool" name="minibooneGA">  false </param>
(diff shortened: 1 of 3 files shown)


dytman:
please keep in mind that dipole approximation for a nucleon->resonance transition is terrible, in fact worse than what Rein-Sehgal provides.  I need to know more about what Afro did before this can be approved.

jtenavidal:
The code has been tested. The results are sumarised here: https://genie-docdb.pp.rl.ac.uk/DocDB/0003/000355/001/EMRES_impact.pdf

jtenavidal:
Steve Dytman suggested to plot the impact of this change against e-N inclusive data: https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ProcessDocumentAdd

It can be seen that this change affects the inclusive cross section prediction. 

It is important to keep in mind that this pull request only changes the configuration - it does not affect the code itself.

jtenavidal:
Correct link: [
https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=362](url)

mroda88 (approved):
LGTM and it's tested.

PR #241, opened 2022-12-05
fix in HA random number usage

mroda88:
This implements exactly what suggested in https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=330


mroda88:
[comparison_hAFix_MinervaTune.pdf](https://github.com/GENIE-MC/Generator/files/10361451/comparison_hAFix_MinervaTune.pdf)

Here are check plot we did.

PR #257, opened 2023-03-02
Sbn/configuration

mroda88:


mroda88:
Notes from discussion with Steven Gardiner. 
 - Remove EM process from tune name
 - Rename tune directory 
 - No need for EventGenerator.xml because it's only there for SuSav2 1p1h

sjgardiner:
Superseded by pull request #258

PR #255, opened 2023-02-22
MARLEY-based nuclear de-excitations for 40Ar

sjgardiner:
Preliminary calculation for an argon target. Only the leading gamma-ray is currently added to the GENIE event record.

PR #242, opened 2022-12-08
Fix for SuSAv2 2p MEC

alons126:
details are in the following document:
[https://genie-docdb.pp.rl.ac.uk/DocDB/0003/000354/001/Pres%2003%20-%20Fixing%20SuSAv2.pdf](url)

PR #258, opened 2023-03-04
v3.4 release candidate

sjgardiner:
This pull request combines commits from pull requests #242 (SuSAv2 EM MEC fixes), #255 (carbon + argon de-excitation gamma-rays), and #257 (new SBN configuration XMLs) in a harmonious way. It supersedes those pull requests and represents all changes needed to tag R-3_04_00.


mroda88 (commented):
That's great. So I'll close the PR for the configuration in favour of this one. 
We need to change the name of the tune. I'll try to do it tomorrow.

mroda88 (approved):
This looks very good now. 
We need to run splines and tests

src/Physics/HadronTensors/HadronTensorI.h
```diff
@@ -42,6 +42,8 @@ typedef enum HadronTensorType {
   kHT_MEC_DeltaAll,
   kHT_MEC_Deltapn,
   kHT_MEC_EM,
+  kHT_MEC_EM_pn,
```
nusense, line 45: Is the HadronTensorType ever saved to any output?   If it is then it's generally not a good idea to add new enum's in the middle of the list.   If it's only ever used internally to the code then it's probably fine-ish.

sjgardiner: Good catch, @nusense! Yeah, it's only used as an internal representation, so I think we can just shrug and move on. But I agree with the "best practices" advice wholeheartedly.

PR #260, opened 2023-03-06
FNAL Scripts Update

jtenavidal:
This is a place holder for further developments on the FNAL scripts 
Problems adressed:
- setting a minimum number of events of 400k for e scattering to avoid high number of jobs runing
- Fixing a bug : GB -> MB for disk space


jtenavidal:
Hi Marco, 
These are the final changes needed to run jobs at the grid. I updated some parameters, this should be all of it now.

jtenavidal:
Adding possibility to pass total xsec file directly. The user doesn't need to create the fake directory structure

PR #262, opened 2023-03-07
BeamHNL bugfix in N4 --> v e mu mode Majorana vs Dirac

kjplows:
Majorana and Dirac decay widths for N4 --> v e mu were switched in BeamHNL/HNLBRCalculator.cxx


mroda88 (approved):
LGTM

PR #277, opened 2023-04-27
Fix Nuclear tune README

jtenavidal:
The readme was pointing out to the wrong experiment. The tunes were performed with MiniBooNE data


mroda88 (approved):
great, thanks for the fix

PR #278, opened 2023-05-02
Add Power Law spectrum

pochoarus:
New feature to generate events with a power law spectrum. Also, the option to force the interaction of all injected events has been included


src/Apps/gEvGen.cxx
```diff
@@ -87,6 +91,11 @@
               scheme for the generated kinematics of individual processes can
               still be in effect if enabled..
               ** Only use that option if you understand what it means **
+           --force
+              Forces the interaction of all injected events.
```
nusense, line 95: I'm not loving the use of "injected events" here.  The term "event" is too overloaded, but to me it generally means an 'interaction'.  Perhaps just "Forces the interaction of all flux rays"

I'm also not clear what it means "if a neutrino flux is specified".  Doesn't one always need to specify some kind of flux?  Do you mean for all cases other than mono-energetic (in which case the mono flux acts as if force was alway in play)?

nusense (changes requested):

src/Apps/gEvGen.cxx
pochoarus: I have changed the first comment accordingly.

Regarding the second comment. You are right, for monoenergetic it is always forced.

src/Tools/LinkDef.h
```diff
@@ -0,0 +1,52 @@
+#ifdef __CINT__
```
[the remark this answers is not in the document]
pochoarus: Indeed, this was a typo. Now it is fixed.

src/Apps/gEvGen.cxx
```diff
@@ -782,6 +816,8 @@ void GetCommandLineArgs(int argc, char ** argv)
        << "Flux: " << gOptFlux;
   LOG("gevgen", pNOTICE)
        << "Generate weighted events? " << gOptWeighted;
+  LOG("gevgen", pNOTICE)
+       << "Force interaction of all injected events? " << gOptForceInt;
```
nusense, line 820: "injected events" --> "flux rays"

src/Apps/gEvGen.cxx
```diff
@@ -77,6 +78,9 @@
                  by default variable bin histograms are multiplied by bin width
                  if ,NOWIDTH then they are not
                  if ,WIDTH then they are alway multiplied by bin width
+              -- A power law:
+                 The general syntax is `-f POWERLAW:alpha'
+                 and the scpectrum will follow a E^{-alpha} distribution
```
nusense, line 83: scpectrum -> spectrum

nusense (approved):
minor changes suggested, but otherwise I think it's fine

pochoarus:
Thanks for the review @nusense . @candreop Any further suggestions before we merge?

pochoarus:
@candreop I think now it is more clear. What do you think?

candreop:
looks good. Thanks @pochoarus

PR #286, opened 2023-06-06
Change GST format to store photons from pi0 decay

jtenavidal:
By default, GENIE doesn't decay pi0's. However, it is possible for the user to decay them. In such cases, the photons are not stored in the output gst, but the pi0 are. This pull request addresses this.


jtenavidal:
Before the change 
image

jtenavidal:
After the change
image

mroda88 (approved):
It was discussed in the forum on 21 June 2023. So we can merge.

PR #288, opened 2023-06-08
Small fix on CRPA implementation

jtenavidal:
The new implementation of CRPA was failing to load the EM hadron tensors due to a small typo in the code. As a consequence, a warning was printed, and the cross section was 0. 
@sjdolan suggested the change.


sjdolan:
Yep, this was a stupid typo. We ran plenty of validations for neutrinos but not for electrons.

Issue #282, opened 2023-05-18
Incorrect xsec_max in RES importance sampling


physcxia:
In the importance sampling of RESKinematicsGenerator, the maximal cross section ``xsec_max`` is being passed to the envelope function in the following line:

https://github.com/GENIE-MC/Generator/blob/25df35f9ca97b6eeab79308b6b49974ffb01d4ef/src/Physics/Resonance/EventGen/RESKinematicsGenerator.cxx#L187

The envelope represents a distribution of the transformed kinematic variable $Q_D^2$. However, the variable ``xsec_max`` is calculated as the maximal differential cross section, $\frac{d\sigma^{\max}}{dWdQ^2}$, with original $Q^2$,

https://github.com/GENIE-MC/Generator/blob/25df35f9ca97b6eeab79308b6b49974ffb01d4ef/src/Physics/Resonance/EventGen/RESKinematicsGenerator.cxx#L104

which has incorrect dimension and numerically may be much less than the maximal differential cross section in $Q_D^2$, $\frac{d\sigma^{\max}}{dWdQ_D^2}$, as expected by the envelope.

The phase space ``kPSWQD2fE`` instead of ``kPSWQ2fE`` should be used in ``ComputeMaxXSec``,

https://github.com/GENIE-MC/Generator/blob/25df35f9ca97b6eeab79308b6b49974ffb01d4ef/src/Physics/Resonance/EventGen/RESKinematicsGenerator.cxx#L370

Could someone help investigate this issue?

candreop:
Thanks for the report, Chen. This is somewhat old code, so we will need to refresh our memory about how the importance sampliing envelope was constructed and investigate further. We will try to do this asap. In the mean time, if you are concerned that the actual differential cross-section overshoots the importance sampling envelope and, therefore, the generation of kinematics distorts the inputs cross-section modek, you can use the safety check implemented in KineGeneratorWithCache::AssertXSecLimits() - See https://github.com/GENIE-MC/Generator/blob/25df35f9ca97b6eeab79308b6b49974ffb01d4ef/src/Physics/Common/KineGeneratorWithCache.cxx.
You can enable this check by setting the MaxXSec-DiffTolerance in https://github.com/GENIE-MC/Generator/blob/master/config/RESKinematicsGenerator.xml

candreop:
**Follow up by Chen Xia:**

[...]

PR #291, opened 2023-06-10
fix RESKinematicsGenerator

igor144:
Fixing of the problem with `RESKinematicsGenerator` reported in issue #282 .

I suggest getting rid of the variable `QD2` to avoid confusion with Jacobian, variable transformation and global maximum searching. The function `double genie::utils::kinematics::RESImportanceSamplingEnvelope(double * x, double * par)` is used only in _RESKinematicsGenerator.cxx_, therefore it should not cause problems anywhere else. All changes are trivial.

### DIFF
--- a/src/Framework/Utils/KineUtils.cxx
+++ b/src/Framework/Utils/KineUtils.cxx
@@ -1373,7 +1373,7 @@ double genie::utils::kinematics::RESImportanceSamplingEnvelope(
                                                      double * x, double * par)
 {
   //-- inputs
-  double QD2   = x[0];   // QD2 (Q2 transformed to take out the dipole form)
+  double Q2    = x[0];   // Q2
   double W     = x[1];   // invariant mass
 
   //-- parameters
@@ -1424,8 +1424,8 @@ double genie::utils::kinematics::RESImportanceSamplingEnvelope(
      }
   }
 
-  if(QD2<0.5) {
-    func *= (1 - (0.5-QD2));
+  if(Q2>controls::kMQD2) {
+    func *= (0.5 + 1./(1 + Q2/controls::kMQD2));
   }
 
   return func;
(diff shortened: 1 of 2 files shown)


mroda88 (commented):
The code looks good to me, have we done some testing and validation of the new generation? 
Can we post it here?

candreop:
Hi Igor: the transformation from Q2 to Q2_d = 1/(1+Q2/c), was done to flatten the differential cross-section by taking out the dipole form. It is easy to remove the transformation and the code would work. But what is the impact on efficiency? If it now becomes much slower, perhaps we want to keep it and fix what is wrong with the sampling envelope? But we need some assessment of CPU efficiency first. Please can you check how fast you can generate RES events w/ and w/o that transformation?

igor144:
> Hi Igor: the transformation from Q2 to Q2_d = 1/(1+Q2/c), was done to flatten the differential cross-section by taking out the dipole form. It is easy to remove the transformation and the code would work. But what is the impact on efficiency? If it now becomes much slower, perhaps we want to keep it and fix what is wrong with the sampling envelope? But we need some assessment of CPU efficiency first. Please can you check how fast you can generate RES events w/ and w/o that transformation?

Hi Costas! I understand your point about flattering, but straightforward comparison of previous version and proposed fix would not be quite right, because in the previous version of code some of the events are accepted, although they should have been rejected. Anyway, such comparison shows, that without fix time/event: 0.00597427 s, with fix: 0.0499688 s.

igor144:
Now, the maximum is searched in W, QD2 variables. We decided to get rid of envelope. 
The test runs of the command:
`time gevgen -p 14 -t 1000010010 -n 10000 -e 0.1,100 -f 1 --cross-sections out.xml --tune G00_00a_00_000 --event-generator-list RES`
showed that it works faster with this option, see  [Presentation.pdf](https://github.com/GENIE-MC/Generator/files/11803250/Presentation.pdf).

mroda88 (approved):
Approved by Costas in the Core developer meeting of the 7 July 2023

PR #292, opened 2023-06-12
Dynamic coupled channels model by Toru Sato

idkakorin:
The `DCCSPPPXSec` and `DCCSPPXSec` algorithm are implemented. They are working for neutrinos and charged leptons.

Now, the following commands are valid:
`gmkspl -p 11L -t 1000000010 -n 30 -e 100 -o out.xml --tune DCC22_00a_00_000` --- calculates cross section for massless left polarized electron on neutron;
`gmkspl -p 11R -t 1000000010 -n 30 -e 100 -o out.xml --tune DCC22_00a_00_000` --- calculates cross section for massless right polarized electron on neutron;
`gmkspl -p 11 -t 1000000010 -n 30 -e 100 -o out.xml --tune DCC22_00a_00_000` --- calculates cross section for massive electron on neutron;
`gmkspl -p 12 -t 1000000010 -n 30 -e 100 -o out.xml --tune DCC22_00a_00_000` --- calculates cross section for electron neutrino on neutron;

The helicity of massless lepton is added to the class `genie::InitialState`.

`DCCSPPPXSec:XSec(const Interaction * i, KinePhaseSpace_t  k) ` calculates:
$\frac{d^2\sigma}{dWdQ^2}$                                           when k = kPSWQ2fE;
$\frac{d^3\sigma}{dWdQ^2d\cos\theta_\pi}$                when k = kPSWQ2ctpfE;
$\frac{d^4\sigma}{dWdQ^2d\cos\theta_\pi d\phi_\pi}$ when k = kPSWQ2ctpphipfE,
where $\theta_\pi$ and $\phi_\pi$ are polar and azimuthal angles of final pion in the resonance rest frame.

The single pion production reaction is defined by enumeration in `SppChannel.h` which was extended for charged lepton case.


mroda88 (dismissed):
There is a small logic mistake. 

I would also ask to add a markdown file in 
```
data/evgen/dcc/
```
to describe the content of the file: units, what it is, link to papers/authors of the file etc. Whatever information we can about it.

src/Framework/Interaction/Interaction.cxx
```diff
@@ -253,6 +253,12 @@ string Interaction::AsString(void) const
   const Target & tgt = fInitialState->Tgt();
 
   ostringstream interaction;
+  
+  char c_hel = 0;
+  if (fInitialState->ProbePdg() == -1)
```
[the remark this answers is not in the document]
igor144: Fixed.

src/Framework/Interaction/Interaction.cxx
```diff
@@ -253,6 +253,12 @@ string Interaction::AsString(void) const
...
+    c_hel = 'L';
+  else if (fInitialState->ProbePdg() == 1)
```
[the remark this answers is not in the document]
igor144: Fixed.

idkakorin (commented):
Address all requests

idkakorin (commented):
Address all requests

config/DCC22_00a/ModelConfiguration.xml
```diff
@@ -0,0 +1,158 @@
...
+
+ </param_set>
+ 
+ <param_set name="Validation">
+    <!--
+	~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	GENIE validity range
+    -->
+    <param type="double" name="GVLD-Emin">    0.010  </param>
+    <param type="double" name="GVLD-Emax">    1.000  </param>
```
nusense: I read this as the CMC can _not_ be used above 1 GeV.   Is that the actual intent?

config/DCC22_00a/TuneGeneratorList.xml
```diff
@@ -0,0 +1,57 @@
...
+Generator-%d                alg     No
+
+-->
+
+  <param_set name="Default"> 
+     <param type="int" name="NGenerators">   3                                    </param>
+     <param type="alg" name="Generator-0">   genie::EventGenerator/DCCSPP-EM      </param>
+     <param type="alg" name="Generator-1">   genie::EventGenerator/DCCSPP-CC      </param>
+     <param type="alg" name="Generator-2">   genie::EventGenerator/DCCSPP-NC      </param>
+  </param_set>
```
nusense, line 53: If this tune only involves DCCSPP it sort of ignores the caution at the top of this file warning against not using a comprehensive configuration, does it not?   

Might it be wiser to have a comprehensive configuration for `Default` and a second configuration for `DCCSPOnly` or such?

src/Apps/gMakeSplines.cxx
```diff
@@ -363,16 +397,49 @@ void PrintSyntax(void)
...
 //____________________________________________________________________________
-PDGCodeList * GetNeutrinoCodes(void)
+PDGCodeList * GetNeutrinoCodes(vector<int> & probe_helicities)
 {
   // split the comma separated list
   vector<string> nuvec = utils::str::Split(gOptNuPdgCodeList,  ",");
 
   // fill in the PDG code list
-  PDGCodeList * list = new PDGCodeList;
+  PDGCodeList * list = new PDGCodeList(true);
```
nusense, line 406: Okay, so this `true` allows for duplicate entries in the PDGCodeList to accommodate the L vs. R felicities for charged probes.   ~~This does leave open the possibility of the user accidentally duplicating the neutral leptons in their command line argument list.~~
I see that preventing duplicates of {pdg,helicity} is handled at the insert level and neutral probes (neutrinos) always are entered as helicity=0.

src/Framework/Interaction/KPhaseSpace.cxx
```diff
@@ -997,31 +981,67 @@ Range1D_t KPhaseSpace::TLim(void) const
...
+double KPhaseSpace::Threshold_SPP(bool isMassless) const
+{
+  const InitialState & init_state = fInteraction->InitState();
+  SppChannel_t spp_channel  = SppChannel::FromInteraction(fInteraction);
+  PDGLibrary * pdglib = PDGLibrary::Instance();
+  double mpi  = pdglib->Find( SppChannel::FinStatePion    (spp_channel) )->Mass();
+  double Mi   = pdglib->Find( SppChannel::InitStateNucleon(spp_channel) )->Mass();
+  double Mf   = pdglib->Find( SppChannel::FinStateNucleon (spp_channel) )->Mass();
+ 
+  double mi = 0, mf = 0;
```
nusense: I'm not a big fan of variables with the same name and differing by just case, e.g. `Mi` vs `mi`.  It leave potential for confusion.   Maybe `mi_p` or some such.  I see that some of the old code also used case to distinguish variables, but I don't like that in retrospect either.

src/Framework/Interaction/KPhaseSpace.cxx
```diff
@@ -91,25 +91,9 @@ double KPhaseSpace::Threshold(void) const
...
-       pion_pdgc = kPdgPiM;
-    else if ( xcls.NPi0() != 1 )
-       throw genie::exceptions::InteractionException("Can't compute threshold");
-    double mpi   = PDGLibrary::Instance()->Find(pion_pdgc)->Mass();
-    double mi    = PDGLibrary::Instance()->Find( init_state.ProbePdg() )->Mass();
-    double mf = ml;
-    double mtot = Mf + mf + mpi; // total mass of FS particles
-    double Ethresh = (mtot*mtot - Mi*Mi - mi*mi)/2/Mi;
-    return Ethresh;
+      return this->Threshold_SPP();
```
nusense, line 94: I like the idea of pushing this computation into a separate function, I'm just having difficulty seeing whether `Threshold_SPP()` gives the same results as what it replaces here.   There's a difference in how it finds `mpi`, `Mi` and `Mf` that looks similar but now uses `SppChannel::X()`.   I'm guessing that the `SppChannel::X()` routines are doing what some of the if/then/else cases are doing here.

src/Framework/Interaction/KPhaseSpace.cxx
```diff
@@ -997,31 +981,67 @@ Range1D_t KPhaseSpace::TLim(void) const
...
+  if (!isMassless)
+  {
+      mi = PDGLibrary::Instance()->Find( init_state.ProbePdg() )->Mass();
+      mf = fInteraction->FSPrimLepton()->Mass();
+  }
+
+  double mtot = Mf + mf + mpi; // total mass of FS particles
+  double Ethresh = (mtot*mtot - Mi*Mi - mi*mi)/2/Mi;
+  return Ethresh;
+}
```
nusense, line 1003: Overall this routine seems more correct w/ fewer assumptions and more compact than what it's replacing.

src/Framework/Interaction/KPhaseSpace.cxx
```diff
@@ -997,31 +981,67 @@ Range1D_t KPhaseSpace::TLim(void) const
...
+  
+  double mi = 0, mf = 0;
+  if (!isMassless)
+  {
+      mi = PDGLibrary::Instance()->Find( init_state.ProbePdg() )->Mass();
+      mf = fInteraction->FSPrimLepton()->Mass();
+  }
+  
+  double Ei   = init_state.ProbeE(kRfHitNucRest);
+  double ECM  = TMath::Sqrt(Mi*(Mi + 2*Ei) + mi*mi);
```
nusense: Does this calculation of ECM differ from the `init_state.CMEnergy()` that it replaces?  The calculation seems to be quite different, but are they equivalent?  Or is the difference in the potential for the probe being massless (with the default being not massless)?

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
```diff
@@ -0,0 +1,1160 @@
+//____________________________________________________________________________
+/*
+ Copyright (c) 2003-2022, The GENIE Collaboration
```
nusense: This code is getting committed in the year 2026, you might as well have the copyright reflect that.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
```diff
@@ -0,0 +1,1160 @@
...
+{
+
+}
+//____________________________________________________________________________
+double DCCSPPPXSec::XSec(const Interaction * interaction, KinePhaseSpace_t kps) const
+{
+
+    // dimension of kine phase space
+    std::string s = KinePhaseSpace::AsString(kps);
+    int kpsdim = s!="<|E>"?1 + std::count(s.begin(), s.begin()+s.find('}'), ','):0;
```
nusense: We should update `KinePhaseSpace::AsString()` to do "the right thing" in the cases:
<pre>
  125 // TODO: update this string when the appropriate kinematic variables are known
  126       case(kPSQELEvGen)   : return "<QELEvGen>"; break;
  127       case(kPSDMELEvGen) : return "<DMELEvGen>"; break;
</pre>
which this code doesn't protect against (no `}`), but presumably are never passed into it.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
```diff
@@ -0,0 +1,1160 @@
...
+    SppChannel_t spp_channel = SppChannel::FromInteraction(interaction);
+
+    int  probepdgc = init_state.ProbePdg();
+    int  nucpdgc   = target.HitNucPdg();
+    int  helicity  = init_state.ProbeHelicity();
+    bool is_nubar  = pdg::IsAntiNeutrino(probepdgc);
+    bool is_EM     = proc_info.IsEM();
+    bool is_CC     = proc_info.IsWeakCC();
+    bool is_NC     = proc_info.IsWeakNC();
+    bool is_EM0    = is_EM && (helicity != 0);
```
nusense, line 107: This seems mislabeled.  It's true if the helicity is _not_ zero, but the name, to me,  implies the opposite.

Is this compatible with line 106 below?  I _don't_ think so, but perhaps I'm confused.  Also used at lines 427, 448 and beyond

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
```diff
@@ -0,0 +1,1160 @@
...
+                val = TMath::Sqrt(2);
+                break;
+            default:
+                val = 0;
+        }
+
+    double sgn = -1;
+    const InitialState & init_state = interaction -> InitState();
+    if (pdg::IsAntiNeutrino (init_state.ProbePdg()))
+        sgn = 1;
```
nusense, line 976: Does this `sgn` need to also change depending on the sign of the charged lepton if that's the probe?  Or is the sign right for electrons and one assumes there's never a positron probe.   Is this code also supposed to hold for muons?

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
```diff
@@ -0,0 +1,1160 @@
...
+    const KPhaseSpace& kps = interaction->PhaseSpace();
+
+    // Get kinematical parameters
+    const InitialState & init_state = interaction -> InitState();
+    const Kinematics & kinematics   = interaction -> Kine();
+    const ProcessInfo &  proc_info  = interaction -> ProcInfo();
+    
+    bool is_EM     = proc_info.IsEM();
+    int  helicity  = init_state.ProbeHelicity();
+    bool is_EM0    = is_EM && (helicity != 0);
```
nusense, line 1278: Again, to me the name `is_EM0` would imply that the it's EM and the helicity is zero, but that's not what the code does.

nusense (commented):
I've made a few comments and asked for a couple clarifications.   Sorry this took so long to review.

config/DCC22_00a/ModelConfiguration.xml
idkakorin: I think I set it for test purpose and then forgot to return back. 
The model restrictions are defined in terms of different variables, namely:
1.077 ≤ W ≤ 2.1
0 ≤ Q² ≤ 3.

config/DCC22_00a/TuneGeneratorList.xml
idkakorin: I agree that there is an inconsistency with the comment regarding the Default configuration. As I mentioned earlier, SPP models such as DCC or MK are currently intended to operate in a stand-alone mode, since work is still required to seamlessly interface them with the RES and DIS models.

For this reason, I kept the default configuration with all these generators so that users do not get the misleading impression that these models are already ready for full use. The same approach is used for the MKSPPP configuration as well.

nusense: okay, I can see that.

src/Apps/gMakeSplines.cxx
idkakorin: Yes, exactly.

src/Framework/Interaction/KPhaseSpace.cxx
idkakorin: Yes, that's correct. The SppChannel::X() helpers encapsulate the same channel-dependent logic that was previously implemented through the if/else blocks.

src/Framework/Interaction/KPhaseSpace.cxx
idkakorin: The function pi.IsSinglePion() was originally introduced for the MK implementation, and it have now been moved it into a separate piece of code that is functionally equivalent to what it replaces here.

src/Framework/Interaction/KPhaseSpace.cxx
idkakorin: They are essentially equivalent. However, you are right — it is better to use init_state.CMEnergy().

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: I agree, will chage it.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: Agreed. I would suggest not updating the list in KinePhaseSpace::AsString(), but instead slightly modifying the code that defines the dimensionality of the phase space.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: The EM0 label refers to electromagnetic processes where the masses of the charged leptons can be neglected (i.e., m = 0). In other words, the “0” refers to the particle mass, not to its helicity. For such effectively massless particles, the helicity can be defined.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: In the original code the sign changes only for antineutrinos in CC events:
https://github.com/GENIE-MC/Attic/blob/master/toru_sato/code-20230612/nun-pids-20230612.f#L863.

Yes, the code is intended to apply to both muons and taus as well.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: The answer is the same as in https://github.com/GENIE-MC/Generator/pull/292#discussion_r2923521939.

src/Framework/Interaction/KPhaseSpace.cxx
idkakorin: I changed variable names.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
nusense: Okay, since the `is_EM0` variable is constructed using the helicity I assumed that the 0 referred to the helicity and not the mass.  Perhaps a comment to that effect?

src/Physics/Resonance/XSection/MKSPPPXSec2020.cxx
```diff
@@ -101,8 +101,15 @@ double MKSPPPXSec2020::XSec(const Interaction * interaction, KinePhaseSpace_t kp
...
-  int kpsdim = s!="<|E>"?1 + std::count(s.begin(), s.begin()+s.find('}'), ','):0;
-  if (kpsdim < 3 || kpsdim > 4) return 0;
+  int kpsdim = 0;
+  if (s != "<|E>") 
+  {
+      std::string::size_type r = s.find('}');
+      if (r != std::string::npos) 
+          kpsdim = 1 + std::count(s.begin(), s.begin() + r, ',');
+  }
+  if (kpsdim < 2 || kpsdim > 4) return 0.;
```
nusense: The prior case had `if (kpsdim &lt; 3 || kpsdim &gt; 4) return 0;` (i.e. only non-zero for 3 or 4) the new code has `if (kpsdim &lt; 2 || kpsdim &gt; 4) return 0.;`  (ie. non-zero for 2, 3, and 4).   I'm assuming that change was intentional and the old code was mistaken.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
nusense: Okay, but the asymmetry here bugs me.  Why not charged anti-particles?

nusense (approved):
Just a couple of comments.   But essentially all my concerns have been addressed so I'm going to approve this PR.

src/Physics/Resonance/XSection/MKSPPPXSec2020.cxx
idkakorin: Yes, you are absolutely right. This was an unintended mistake introduced on my side due to a copy-paste. MK does not produce two-dimensional cross sections, so the previous condition (non-zero only for 3 or 4) was the correct one.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: Good point — I’ll add a comment to make that explicit.

src/Physics/Resonance/XSection/DCCSPPPXSec.cxx
idkakorin: Good question. I’ll try to clarify this with the model author.

idkakorin:
Following the request from Steven Gardiner (@sjgardiner) and the suggestions from Steve Dytman (@dytman), I have prepared validation plots based on comparisons with electron scattering data. 

These results were produced using a modified version of the Comparisons framework (branch develop/eNFit), adapted specifically for this purpose.

Four sets of results were generated:

1. DCCSPP and Berger-Sehgal with DIS
2. DCCSPP with DIS (χ² = 6.25927e+10 / 12159 DoF)
3. DCCSPP without DIS (χ² = 2.25324e+09 / 12159 DoF)
4. Berger-Sehgal with DIS (χ² = 4.12735e+10 / 12159 DoF)

These comparisons provide a first validation of the DCC SPP implementation in the absence of a full event generator.
[genie-e_res_data_comp-BS_with_DIS.pdf](https://github.com/user-attachments/files/26159910/genie-e_res_data_comp-BS_with_DIS.pdf)
[genie-e_res_data_comp-DCC.pdf](https://github.com/user-attachments/files/26159911/genie-e_res_data_comp-DCC.pdf)
[genie-e_res_data_comp-DCC_vs_BS_with_DIS.pdf](https://github.com/user-attachments/files/26159912/genie-e_res_data_comp-DCC_vs_BS_with_DIS.pdf)
[genie-e_res_data_comp-DCC_with_DIS.pdf](https://github.com/user-attachments/files/26159913/genie-e_res_data_comp-DCC_with_DIS.pdf)

idkakorin:
Add extra comparison set: DCC (with DIS) vs Bosted-Christy Fit vs Berger-Sehgal (with DIS).

[genie-e_res_data_comp-DCC_vs_BS_with_DIS_vs_BoostedChristy.pdf](https://github.com/user-attachments/files/26679933/genie-e_res_data_comp-DCC_vs_BS_with_DIS_vs_BoostedChristy.pdf)

PR #296, opened 2023-06-19
Remove fatal error when a TGeoMixture object reports a duplicate nuclear PDG code

sjgardiner:
This is no longer necessarily an error condition due to changes in how ROOT v6.25.02 handles isomers. See https://github.com/root-project/root/pull/8556 for details.


nusense (approved):

candreop:
Don't we need a 
#if ROOT_VERSION_CODE >= ROOT_VERSION(6,25,2)
#endif
block?

sjgardiner:
@candreop A check of the ROOT version seems like overkill to me here, but @nusense should please give his opinion. Even with older versions of ROOT, throwing this fatal error may not be as useful as it originally seemed to be. In any case, the code will never return an incorrect weight (it sums over contributions from duplicate PDG codes as needed).

nusense:
No, I don't think the ROOT_VERSION conditional is necessary.  I'm not sure why it was ever a fatal error in the first place -- just that it never happened with the previous version of TGeoMixture handling.  I'm not sure how it could ever be triggered by the previous ROOT versions.

PR #299, opened 2023-07-03
removing EM parameter for Nieves. Model only for neutrinos

jtenavidal:
There was a Scaling EM parameter for Nieves but this model doesn't support EM mode. I removed the parameter from the implementation


mroda88 (approved):
LGTM

PR #301, opened 2023-07-08
fix compilation errors of QPMDMDISPXSec

igor144:
1. Fix errors which occurs at compilation of class `QPMDMDISPXSec`.
2. Change priority of **XSecSplLst** from WARN (set by Alfonso) to NOTICE in _Messenger.xml_, in other case **gmkspl** doesn't display anything.
3. Fix a typo in _Spline.cxx_.


mroda88 (approved):
LGTM

PR #302, opened 2023-07-09
Fix edge issue in GMCJDriver

igor144:
Now, the maximum energy of the spline built inside the `GMCJDriver` is defined as the maximum of validity energy range for this generator. Sometimes there are cases when the splines are pre-calculated in a range whose maximum is lower than the maximum of the validity energy range. In such cases, the total cross section spline is incorrect at the highest energies and the generated events are skewed.


mroda88 (approved):
LGTM

PR #303, opened 2023-07-18
add ability to specify flux direction in gEvGen.cxx

nusense:


nusense:
Turns out the "radius" and "spot" settings are irrelevant when using the `PointGeomAnalyzer` that `gevgen` uses -- this always puts the vertex at (0,0,0).

candreop:
Can we add the new settings to the manual? I'll send the link to editable version of the Overleaf document at our Slack channel.

candreop (approved):

PR #306, opened 2023-07-26
Addressing EM RES issue

jtenavidal:
The new RES-EM Scaling parameter was not loaded in the BS xsec class. As a consequence, it was given a very small number, causing the xsec calculation to be always 0. This PR fixes this. I had a close look at the other processes. This behaviour is not observed, indicating the only problematic process was EMRES for the BS model.


jtenavidal:
I also took the chance to fix a small typo in one printout. Instead of GV it should say it it is "Using dipole parametrization for GA"

mroda88 (approved):
ok

PR #307, opened 2023-07-26
Develop/fsi setup fnal

jtenavidal:
This includes the possibility for the user to run GENIE with G4 or INCl at the grid. The user doesn't have to do anything else but calling the corresponding tunes. The scripts will deal with the rest.


mroda88 (approved):
LGTM

PR #309, opened 2023-07-28
Fix EM-XSec factor, delete unused variables in BSKLNRES model

igor144:
1. EM factor wasn't calculated appropriately - fix it.
2. Due to the first, there is no need to keep sinus of Weinberg angle.


mroda88 (approved):
LGTM

jtenavidal (commented):
LGTM

PR #323, opened 2023-09-07
Config cleanup

mroda88:
While working on the TKI tuning, we realised some INUKE related entries in `CommonParam.xml` files were completely obsolete and created confusion.

Comments were moved in `H*Intranuke*.xml` and the blocks were removed as they were not even read from the code.


feuerbutter (approved):
I checked and all are removing the correct segment.

feuerbutter (approved):
I checked and all are removing the correct segment

PR #330, opened 2023-09-26
Electron distribution

mroda88:


src/Physics/NuElectron/XSection/PXSecOnElectron.cxx
```diff
@@ -0,0 +1,118 @@
...
+  int NInt = 0; //Count number of integrations
+  do{
+    NInt++;
+    Interaction in_curr(*interaction); //Copy interaction object
+    fElectronVelocity->InitializeVelocity(in_curr); //Modify interaction to give electron random velocity from selected distribution
+    double xsec = fXSecIntegrator->Integrate(this,&in_curr); // In ele at rest
+    //get beta - comps orthogonal to beam x,y
+    //scale = sqrt(1-b_t^2)
+    TVector3 beta = in_curr.InitState().Tgt().HitPartP4().BoostVector(); // beta
+    double beta_tangent = sqrt(TMath::Power(beta[0],2)+TMath::Power(beta[1],2)); //Component tangential to beam
```
mroda88, line 63: This is wrong. What we need here is a the component of beta orthogonal to the probe. Since the probe is not always parallel to z this math needs to be rewritten

bear-is-asleep (approved):

PR #331, opened 2023-09-26
Implement Electron Velocity Changes

bear-is-asleep:


config/BardinIMDRadCorPXSec.xml
```diff
@@ -5,8 +5,24 @@
...
+  NuE-XSecRelError        double   No                                         genie::PXSecOnElectron/Default
+  ....................................................................................
+  -->
+
   <param_set name="Default"> 
-     <param type="alg" name="XSec-Integrator"> genie::IMDXSec/Default </param>
+     <param type="alg" name="XSec-Integrator"> genie::XSecOnElectron/Default </param>
+      <param type="alg"  name="Electron-Velocity"> genie::ElectronVelocityMap/Default  </param>
+    <param type="int"    name ="N-Integration-Samples" > genie::PXSecOnElectron/Default  </param>
+    <param type="double"    name ="NuE-XSecRelError" > genie::PXSecOnElectron/Default  </param>
```
mroda88, line 25: mismatch between type and content

config/IMDAnnihilationPXSec.xml
```diff
@@ -3,16 +3,23 @@
...
+NuE-XSecRelError        double   No                                         genie::PXSecOnElectron/Default
 ....................................................................................
 -->
 
   <param_set name="Default"> 
-     <param type="alg"  name="XSec-Integrator"> genie::IMDXSec/Default  </param>
+    <param type="alg" name="XSec-Integrator"> genie::XSecOnElectron/Default </param>
+    <param type="alg"  name="Electron-Velocity"> genie::ElectronVelocityMap/Default  </param>
+    <param type="int"    name ="N-Integration-Samples" > genie::PXSecOnElectron/Default  </param>
+    <param type="double"    name ="NuE-XSecRelError" > genie::PXSecOnElectron/Default  </param>
```
mroda88, line 22: same comment as before

config/PXSecOnElectron.xml
mroda88, line 1: This file should not exisist as we don't have configurations for interface only

config/StaticElectronVelocity.xml
mroda88, line 1: Since this is empty, we can configure master_config.xml to point to dummy.xml

config/XSecOnElectron.xml
mroda88, line 1: Remove this file

mroda88 (changes requested):
Starting to make comments. Here are a few initial

src/Physics/NuElectron/EventGen/NuEKinematicsGenerator.cxx
mroda88, line 1: Here we have to add something that was not present before. The motion of the electron can cause the process to go under threshold, imagine the electron moving away from the neutrino. In the neutrino-electron elastic scattering that has no threshold this is not possible. But in the IMD this can happen. 

So here before going on with the generation of the event we need to add a check on the ValidKinematics. If that is not the case we have to throw an exception saying that the event is rejected and we have to generate another electron velocity.

mroda88 (changes requested):
Most of the changes are technical. There is one issue with the threshold though.

mroda88:
With this I had a look at the whole thing. On top of the "errors" I pointed out, I think we need a few testing macro, to verify that the events are produced in the right way. 

I think we just need a loop over the events that extracts
 - the distribution of the hit electron
 - the y distribution of the interaction, as we want to prove that the old system and the new one are the same.

src/Framework/Interaction/InitialState.cxx
```diff
@@ -346,9 +346,9 @@ TLorentzVector * InitialState::GetProbeP4(RefFrame_t ref_frame) const
              // make sure that 'struck nucleon' properties were set in
              // the nuclear target object
 
-             assert( fTgt->HitNucP4Ptr() != 0 );
+	     assert( pdg::IsNucleon(fTgt->HitPartPdg()) );
 
-             TLorentzVector * pnuc4 = fTgt->HitNucP4Ptr();
+             TLorentzVector * pnuc4 = fTgt->HitPartP4Ptr();
 
              // compute velocity vector (px/E, py/E, pz/E)
```
bear-is-asleep, line 354: I believe we're missing a reference frame case for striking the electron's reference frame. https://github.com/bear-is-asleep/Generator/blob/255d21798dc6f7382ebe503015535326289fb272/src/Framework/Interaction/InitialState.cxx#L394-L410

bear-is-asleep: same comment for both `GetTgtP4` and `GetProbeP4` https://github.com/bear-is-asleep/Generator/blob/25410911ff9227f5b9eb05a6dc755aad892cb327/src/Framework/Interaction/InitialState.cxx#L322-L339

config/StaticElectronVelocity.xml
bear-is-asleep: I only see `Default.xml` , and I've configured the master_config to point here and deleted this xml

src/Physics/NuElectron/XSection/PXSecOnElectron.h
```diff
@@ -0,0 +1,71 @@
...
+\brief    nu/nubar + e- scattering differential cross section \n
+          The cross section algorithm interfaces nu+e cross sections
+          Final states are handled by sub modules 
+
+          Is a concrete implementation of the XSecAlgorithmI interface. \n
+
+\ref      W.J.Marciano and Z.Parsa, Neutrino-electron scattering theory,
+          J.Phys.G: Nucl.Part.Phys. 29 (2003) 2629-2645
+\ref      Oleksandr Tomalak and Richard J. Hill, Theory of elastic neutrino-electron scattering,
+          Phys. Rev. D 101, 033006 – Published 21 February 2020
```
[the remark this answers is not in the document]
bear-is-asleep: done, and we did not use Tomolak's corrections so I've removed this reference

src/Physics/NuElectron/XSection/NuElectronPXSec.cxx
```diff
@@ -132,33 +142,15 @@ double NuElectronPXSec::XSec(
...
   this->LoadConfig();
 }
 //____________________________________________________________________________
 void NuElectronPXSec::Configure(string config)
 {
-  Algorithm::Configure(config);
+  PXSecOnElectron::Configure(config);
   this->LoadConfig();
 }
 //____________________________________________________________________________
```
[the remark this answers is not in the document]
bear-is-asleep: I removed these functions since they're declared final in `PXSecOnElectron`

src/Physics/NuElectron/XSection/BardinIMDRadCorPXSec.cxx
```diff
@@ -268,10 +256,6 @@ void BardinIMDRadCorPXSec::LoadConfig(void)
   ////fIntegrator =
```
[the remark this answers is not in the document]
bear-is-asleep: i left it for now..

src/Physics/NuElectron/XSection/PXSecOnElectron.cxx
```diff
@@ -0,0 +1,118 @@
...
+    double xsec_err = sqrt((xsec_sum2/NInt-TMath::Power(xsec_mean,2))/NInt)/xsec_mean;
+    if (NInt > 1 && xsec_err < fErrTolerance){break;} //Break condition for dipping below set tolerance
+  }
+  while ( NInt < fNIntegration); 
+  double xsec_avg = xsec_sum/NInt;
+  return xsec_avg;
+}
+//____________________________________________________________________________
+bool PXSecOnElectron::ValidProcess(const Interaction * interaction) const
+{
```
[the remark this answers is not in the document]
bear-is-asleep: We haven't set the target pdg, so waiting to implement this check

src/Physics/NuElectron/EventGen/NuEKinematicsGenerator.cxx
bear-is-asleep: Added this check in `NuEKinematicsGenerator.cxx` - 

```
//Check if kinematics is still valid
  //CM energy may have dropped below threshold due to initial state electron velocity
  if(!fXSecModel->ValidKinematics(evrec->Summary())) {
     LOG("NuEKinematics", pWARN)
          << "The current kinematics configuration is unphysical";
     genie::exceptions::EVGThreadException exception;
     exception.SetReason("Event below threshold");
     exception.SwitchOnStepBack();
     throw exception;
  }
```

src/Physics/NuElectron/XSection/PXSecOnElectron.cxx
```diff
@@ -0,0 +1,118 @@
...
+  int NInt = 0; //Count number of integrations
+  do{
+    NInt++;
+    Interaction in_curr(*interaction); //Copy interaction object
+    fElectronVelocity->InitializeVelocity(in_curr); //Modify interaction to give electron random velocity from selected distribution
+    double xsec = fXSecIntegrator->Integrate(this,&in_curr); // In ele at rest
+    //get beta - comps orthogonal to beam x,y
+    //scale = sqrt(1-b_t^2)
+    TVector3 beta = in_curr.InitState().Tgt().HitPartP4().BoostVector(); // beta
+    double beta_tangent = sqrt(TMath::Power(beta[0],2)+TMath::Power(beta[1],2)); //Component tangential to beam
```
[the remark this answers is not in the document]
bear-is-asleep: I added `GetParallel` and `GetOrthogonal` functions into `Framework/Numerical/MathUtils.h` and modified the boost to be transverse wrt the incoming neutrino

config/BardinIMDRadCorPXSec.xml
mroda88: yes, that simply returns a 0 or a `NaN` it won't do what you want

config/PXSecOnElectron.xml
mroda88: Only the non virtual classes will have their own configuration file

config/BardinIMDRadCorPXSec.xml
```diff
@@ -5,8 +5,24 @@
...
+  NuE-XSecRelError        alg      No         Reference PXSec (double)        genie::PXSecOnElectron/Default
+  ....................................................................................
+  -->
+
   <param_set name="Default"> 
-     <param type="alg" name="XSec-Integrator"> genie::IMDXSec/Default </param>
+    <param type="alg"     name="XSec-Integrator"> genie::XSecOnElectron/Default </param>
+    <param type="alg"     name="Electron-Velocity"> genie::ElectronVelocityMap/Default  </param>
+    <param type="alg"     name ="N-Integration-Samples" > genie::PXSecOnElectron/Default  </param>
+    <param type="alg"     name ="NuE-XSecRelError" > genie::PXSecOnElectron/Default  </param>
```
mroda88, line 25: No these are fixed in the wrong way. These are supposed to be integer and double. The type was correct, you had to fix the value

config/IMDAnnihilationPXSec.xml
```diff
@@ -3,16 +3,23 @@
...
+  NuE-XSecRelError        alg      No         Reference PXSec (double)        genie::PXSecOnElectron/Default
+  ....................................................................................
+  -->
 
   <param_set name="Default"> 
-     <param type="alg"  name="XSec-Integrator"> genie::IMDXSec/Default  </param>
+    <param type="alg"     name="XSec-Integrator"> genie::XSecOnElectron/Default </param>
+    <param type="alg"     name="Electron-Velocity"> genie::ElectronVelocityMap/Default  </param>
+    <param type="alg"     name ="N-Integration-Samples" > genie::PXSecOnElectron/Default  </param>
+    <param type="alg"     name ="NuE-XSecRelError" > genie::PXSecOnElectron/Default  </param>
```
mroda88, line 22: Same problem

src/Framework/Numerical/MathUtils.h
```diff
@@ -128,6 +129,9 @@ namespace math
   double NonNegative    (double x);
   double NonNegative    (float  x);
 
+  TLorentzVector GetParallel(const TLorentzVector lv, const TVector3 unitVec);
+  TLorentzVector GetOrthogonal(const TLorentzVector lv, const TVector3 unitVec);
```
mroda88, line 133: If the inputs are `const`, make them references or don't make them `const` at all if they are copies. 
I prefer the const reference version

src/Framework/Numerical/MathUtils.cxx
```diff
@@ -284,3 +286,20 @@ double genie::utils::math::NonNegative(float x)
...
+    auto orthogonal_mom_3d = lv3d - par_mom;
+    return TLorentzVector(orthogonal_mom_3d, lv.E());
+}
+//____________________________________________________________________________
+TLorentzVector genie::utils::math::GetParallel(const TLorentzVector lv, const TVector3 unitVec) {
+    // Returns TLorentz vector that is parallel to a given unit vector in 3d
+    auto lv3d = lv.Vect();
+    auto par_mom_module = lv3d.Dot(unitVec);
+    TVector3 par_mom = par_mom_module * unitVec;
+    return TLorentzVector(par_mom, lv.E());
```
mroda88, line 304: Looking good, please add a macro that checks if they do the right thing.

src/Physics/NuElectron/XSection/BardinIMDRadCorPXSec.cxx
mroda88, line 1: To make this right you also need to set the energy of the probe to be the one in the hit electron reference frame

src/Physics/NuElectron/XSection/IMDAnnihilationPXSec.cxx
mroda88, line 1: This needs to use the energy in the electron at rest reference frame

src/Physics/NuElectron/XSection/PXSecOnElectron.cxx
```diff
@@ -0,0 +1,128 @@
...
+    //TVector3 beta = in_curr.InitState().Tgt().HitPartP4().BoostVector(); // beta
+    //double beta_tangent = sqrt(TMath::Power(beta[0],2)+TMath::Power(beta[1],2)); //Component tangential to beam
+    
+    auto probe_direction = in_curr.InitState().GetProbeP4(kRfHitElRest)->Vect().Unit();
+    auto electron_mom = in_curr.InitState().Tgt().HitPartP4();
+    auto transverse_mom = utils::math::GetOrthogonal(electron_mom, probe_direction);
+    auto beta_transverse = transverse_mom.BoostVector().Mag();
+    //double beta_transverse = beta_transverse.Mag();
+
+    xsec *= sqrt(1-TMath::Power(beta_transverse,2)); //Correct for lorentz factor
```
mroda88, line 72: Don;t reinvent the wheel, TLorentzVector has a its own gamma function, you can use that

src/contrib/electron_motion/validateNUE.C
mroda88, line 1: Here we need to add the y distributions. And we should have the different distributions for elastic scattering, IMD and IMD annihilation

mroda88 (changes requested):
A few logical errors still

src/Framework/Numerical/MathUtils.cxx
bear-is-asleep: Done

src/contrib/electron_motion/validateNUE.C
bear-is-asleep: there is a kinematic y distribution being made

src/scripts/gcint/loadlibs.C
mroda88, line 1: For these changes we might want to wait for Robert's comment

.gitignore
```diff
@@ -14,3 +14,5 @@ src/Framework/Conventions/GVersion.h
 *.rootmap
 *_ROOT_DICT_*
 *.swp
+src/contrib/electron_motion/Plots
+*.diff
```
mroda88, line 18: I would prefer this not not change

config/BardinIMDRadCorPXSec.xml
```diff
@@ -5,8 +5,24 @@
...
+  <!--
+  Configuration for the BardinIMDRadCorPXSec xsec algorithm.
+
+  Configurable Parameters:
+  ....................................................................................
+  Name                    Type     Optional   Comment                         Default
+  XSec-Integrator         alg      No                                         genie::XSecOnElectron/Default 
+  Electron-Velocity       alg      No                                         genie::ElectronVelocityMap/Default
+  N-Integration-Samples   int      No                                         10000
+  NuE-XSecRelError        double   No                                         0.0001
```
mroda88, line 17: more description here. Explain here that N integration is the maximum number of iteration in case we don't reach the desired relative error

config/IMDAnnihilationPXSec.xml
```diff
@@ -3,16 +3,23 @@
...
+Configuration for the IMDAnnihilationPXsec algorithm.
 
 Configurable Parameters:
 ....................................................................................
-Name           Type     Optional   Comment                 Default
+Name                    Type     Optional   Comment                         Default
+XSec-Integrator         alg      No                                         genie::XSecOnElectron/Default 
+Electron-Velocity       alg      No                                         genie::ElectronVelocityMap/Default
+N-Integration-Samples   int      No                                         10000
+NuE-XSecRelError        double   No                                         0.0001
```
mroda88, line 14: Same as before

config/PXSecOnElectron.xml
```diff
@@ -0,0 +1,28 @@
...
+<alg_conf>
+
+<!--
+Configuration for the PXSecOnElectron xsec algorithm.
+
+Configurable Parameters:
+....................................................................................
+Name                    Type     Optional   Comment                         Default
+N-Integration-Samples   int      No         Maximum integration iterations  10000
+NuE-XSecRelError        double   No                                         0.001
```
mroda88, line 12: Same comment as before

config/PXSecOnElectron.xml
mroda88, line 1: If I'm not wrong, this is only an interface, so we don't need the xml file, unless it's used in master_config.xml to be used for common interfaces.

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+
+      //Print out values
+      std::cout<<"Mass = "<<electron->Mass()<<std::endl;
+      std::cout<<"px = "<<pe.Px()<< " py = "<<pe.Py()<<" pz = "<<pe.Pz()<<std::endl;
+      std::cout<<"vx = "<<vx<< " vy = "<<vy<<" vz = "<<vz<<std::endl;
+      std::cout<<"costheta = "<<costheta<<" phi = "<<phi<<std::endl;
+      
+
+    }//If is scattering
+
```
mroda88, line 93: I fear here we need a
```C++
mcrec -> Clean() ;
```
It's an issue we noticed in recent versions of ROOT.

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+
+  //SEt histograms
+  TH1F *vx_dist = new TH1F("vx",char_title_car,20,-0.15,0.15);
+  TH1F *vy_dist = new TH1F("vy",char_title_car,20,-0.15,0.15);
+  TH1F *vz_dist = new TH1F("vz",char_title_car,20,-0.15,0.15);
+
+  //2D histogram with cos theta and phi of velocities
+  TH2F *ang_dist = new TH2F("vxy",char_title_ang,20,-1,1,20,-3.1415,3.1415);
+
```
mroda88, line 50: I would like a distribution of the momentum module as well.

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+  canvas.Print(fo);
+
+  TCanvas canvas2("canvas2");
+  ang_dist->Draw("COLZ");
+  ang_dist->GetXaxis()->SetTitle("cos#theta");
+  ang_dist->GetYaxis()->SetTitle("#phi");
+  ang_dist->SetStats(kFALSE);
+  std::string file_out2 = "Plots/angdist.pdf";
+  const char *fo2 = file_out2.c_str();
+  canvas2.Print(fo2);
```
mroda88, line 133: Can you add an output file too and add the hists (and the canvas if you like) in the file as well?

mroda88: And make the printing of the canvas optional, most of the time we don't have an X terminal opened.

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+
+  TCanvas canvas("canvas");
+  y_dist->Draw("HIST");
+  y_dist->SetLineColor(kBlack);
+  y_dist->SetFillColorAlpha(kBlue,0.85);
+  y_dist->GetXaxis()->SetTitle("y");
+  y_dist->SetStats(kFALSE);
+  std::string file_out = "Plots/ydist.pdf";
+  const char *fo = file_out.c_str();
+  canvas.Print(fo);
```
mroda88, line 178: Same here, hists stored in a file as well. And optional canvas print.

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+
+void validateNUE(TString filename = "gntp.0.ghep.root",TString dir=""){
+  //Make all distributions
+  gSystem->Exec("mkdir -p Plots");
+  make_vdist(filename,dir);
+  make_y_dist(filename,dir);
+}
+
+void make_finale_tree(TString filename = "gntp.0.ghep.root", TString dir=""){
+  std::vector<int> nu_pdgs = {12,14,-12,-14};
```
mroda88, line 189: Since it's a static vector, use std::array.

Also, what happened to the tau?

src/contrib/electron_motion/validateNUE.C
```diff
@@ -0,0 +1,335 @@
...
+}
+
+void validateNUE(TString filename = "gntp.0.ghep.root",TString dir=""){
+  //Make all distributions
+  gSystem->Exec("mkdir -p Plots");
+  make_vdist(filename,dir);
+  make_y_dist(filename,dir);
+}
+
+void make_finale_tree(TString filename = "gntp.0.ghep.root", TString dir=""){
```
mroda88, line 188: I'm worried that here we are reinventing the wheel, what is this doing differently other than the standard gst format?

src/Physics/NuElectron/XSection/XSecOnElectron.h
```diff
@@ -33,12 +36,12 @@
 
 namespace genie {
 
-class NuElectronXSec : public XSecIntegratorI {
+class XSecOnElectron : public XSecIntegratorI {
```
mroda88, line 39: We need to update also the `#ifdef` guards.

mroda88 (changes requested):
We are getting there. Some changes to be made. 

Let's not forget the ValidKinematic check ow that the HitPartPdg is set correctly. 
We need to understand if the energy threshold is evaluated correctly. in the framework.

config/PXSecOnElectron.xml
bear-is-asleep: deleted this fcl since it's an interface

config/PXSecOnElectron.xml
bear-is-asleep: yep!

src/Physics/NuElectron/XSection/PXSecOnElectron.cxx
```diff
@@ -0,0 +1,126 @@
...
+{
+  bool is_valid = true;
+  if(!pdg::IsElectron(interaction->InitState().TgtPtr()->HitPartPdg())) is_valid = false;
+  if(!interaction->TestBit(kISkipProcessChk)) is_valid = false;
+  return is_valid;
+}
+//____________________________________________________________________________
+bool PXSecOnElectron::ValidKinematics(const Interaction* interaction) const
+{
+  if(interaction->TestBit(kISkipKinematicChk)) return true;
```
mroda88, line 93: We still need to check the energy threshold. I think here should be right place. unless it's done somewhere else.

src/Physics/NuElectron/EventGen/NuEKinematicsGenerator.cxx
```diff
@@ -58,6 +62,17 @@ void NuEKinematicsGenerator::ProcessEventRecord(GHepRecord * evrec) const
   const EventGeneratorI * evg = rtinfo->RunningThread();
   fXSecModel = evg->CrossSectionAlg();
 
+  //Check if kinematics is still valid
+  //CM energy may have dropped below threshold due to initial state electron velocity
+  if(!fXSecModel->ValidKinematics(evrec->Summary())) {
```
mroda88, line 67: This valid kinematics does not make sense if we don't check we are above threshold in the function.

mroda88 (changes requested):
There are still problems with the handling of the threshold. As a matter of fact, we should check if `IsAboveThreshold` from the kinematic phase space does the right job now that we changed the interaction.

PR #332, opened 2023-10-03
Adding missing libraries in the root environment

mroda88:


nusense (approved):
This looks fine to me

PR #339, opened 2023-10-20
Addition of new Vector Form Factors for RES

jtenavidal:
The default GENIE for neutrinos is based on "miniBooNE approach". This approach was incorrect for electrons. The correct method for electrons is added - it is called Sarita-Schwinger. The dipole is also an option. The best agreement to data is shown by Sarita-Schwinger.
[DefaultMBFF.pdf](https://github.com/GENIE-MC/Generator/files/13055659/DefaultMBFF.pdf)
[DipoleFF.pdf](https://github.com/GENIE-MC/Generator/files/13055661/DipoleFF.pdf)
[SRFF.pdf](https://github.com/GENIE-MC/Generator/files/13055662/SRFF.pdf)


jtenavidal:
The default parameterization for neutrinos is using the "MiniBooNE prenscription".

For electrons, we are for now using the Dipole prenscription. 

I had a look at this paper: https://journals.aps.org/prd/pdf/10.1103/PhysRevD.77.053001
They suggest that the Rarita-Schwinger formalism for V+A form factors has better agreement with electron and neutrino data. 

This prenscription is very similar to the current implementation for MiniBooNE. In fact, they should be exact, but they are not. The Rarita-Schwinger provides with specific form for EM or CC/NC, but this differentiation was not implemented. In addition, there's two results for the Axial part (Type I and Type II). Both are available on this new version. 

Status: I am currently running splines for all cases to evaluate the impact on the electron side, and neutrino side.

jtenavidal:
image

jtenavidal:
[comparisonRES.pdf](https://github.com/GENIE-MC/Generator/files/13190323/comparisonRES.pdf)

mroda88 (approved):
Ok

PR #340, opened 2023-10-23
Adding Wmin for Non-RES background

jtenavidal:
We observed that for some tuned eN RES GV FF, the delta peak is well described without non-res background. We add a new parameter, Wcutmin, to cut the region where we believe non-res background is not necessary. The default is set to 0 to not alter the existing configurations. 

If needed, a new parameter can be used to change the coverage of the non-res bkg in W. 

The example attached considers Wmin of 1.4GeV:
[testNewWmin.pdf](https://github.com/GENIE-MC/Generator/files/13069606/testNewWmin.pdf)

The expected behaviour is observed when using default of 0:
[testnew.pdf](https://github.com/GENIE-MC/Generator/files/13069610/testnew.pdf)


jtenavidal:
Best value: Wcutmin=1.25 GeV
[test125.pdf](https://github.com/GENIE-MC/Generator/files/13069684/test125.pdf)

PR #342, opened 2023-10-24
add IBD list for VLE tune

distefanoc:
IBD list has been added in config/EventGeneratorListAssembler.xml to select only this channel running the VLE tune

in src/Physics/NuclearDeExcitation/NucDeExcitationSim.cxx line 413, const has been added defining ns12


mroda88 (approved):
LGTM

PR #351, opened 2023-10-27
Update CommonParam.xml

jtenavidal:
There were some missing parameters in the xml file and the tune couldn't be run.


mroda88 (approved):
OK

PR #353, opened 2023-10-30
PROFRW - Move professor2 from reweight to GENIE

karuboniru:
Professor2 is not yet used directly inside Generator itself, but this
will be useful for the WIP reweight update. Since both comparison
and reweight package is going to be linked to  professor2. We are
putting professor2 as external code for GENIE.

The version selected is professor-2.4.1. The design of submodule 
system only allow specifying branch or commit. 

Checking out of submodules will be done automatically when 
user call `configure` with `--enable-professor2`.


mroda88 (approved):
Great!

PR #361, opened 2023-11-09
Use Singleton Messenger

karuboniru:
Previous implementation of Messenger will deconstruct before some class that are using this in its deconstructor. This will lead to noisy output and a sigsgev error. 

This should fix it and always make sure Messenger is deconstructed at correctly order.


karuboniru:
oops, I think this is not working because it still can't ensure that Messanger is deconstrued after its consumers. This "fixes" the bug just by returning a dangling pointer to deconstructed object that happens to work

PR #362, opened 2023-11-09
Develop/ha direct fate tweak

feuerbutter:
- scaled meanfreepath in Intranuke2018 to allow independent tweaking of each FSI fate. 
- the rescaling in HAintranuke2018 is not necessary as it is throwing the dice in effectively the same way


mroda88 (changes requested):
You created a new CommonParam, that is not necessary and redundant as there is one already for that tune. Please remove it.

PR #366, opened 2023-11-17
Provides ways to load Reweight configurations from AlgConfigPool

karuboniru:
Currently `AlgConfigPool` will only load configuration from `GENIE/config`. While we are planning to make some components from reweight become configurable with  `AlgConfigPool`. This requires some modifications to current  `AlgConfigPool` and related code:

 - Don't early return when failed to load tunning information
 - Try to load `reweight_master_config.xml` in addition to `master_config.xml`
 - Add `GENIE_REWEIGHT/config` to search dirs for confiurations

### DIFF
--- a/src/Framework/Algorithm/AlgConfigPool.cxx
+++ b/src/Framework/Algorithm/AlgConfigPool.cxx
@@ -8,6 +8,7 @@
 */
 //____________________________________________________________________________
 
+#include <cstring>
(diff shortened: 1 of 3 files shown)


mroda88 (approved):
Approved in discussion with Costas. 

We'll wait to merge until the reweight master config file is also merged in reweight.

PR #374, opened 2023-12-02
 add Fermi and fine structure constants in all CommonParams.xml to avoid problem with configuration

idkakorin:
and fix warnings: Unused class rule


jtenavidal:
Yes, I suspect this is the problem too. For me it was only working for untuned configurations. When using tunes, it was breaking it.

mroda88 (approved):
Great!

PR #376, opened 2024-01-19
configure: change detection of Root MathMore

aphecetche:
I'm working on a system (a Mac M1 with Root installed using the [Spack](https://spack.io) package manager) where the Root libraries are installed under `$ROOTSYS/lib/root`, not `$ROOTSYS/lib` so the original detection of whether or not Root was compiled with mathmore is not working. 

I'm proposing here a fix based on the `root-config` executable instead.


nusense (approved):
Looks like a much more robust test than before.

PR #385, opened 2024-02-14
removing conversion as it leads to problematic cases

jtenavidal:
It creates problematic files.


mroda88 (approved):
LGTM because restricted to FNAL scripts

PR #387, opened 2024-02-15
Implement time shift in GSimpleNtpFlux::MoveToZ0()

sjgardiner:
The `GSimpleNtpFlux` class allows the user to shift the starting z coordinate for a neutrino ray off of the flux window. When this is done, the other spatial coordinates are updated accordingly, but the time is set to zero.

The failure to set a consistent shift to the starting time appears to be the cause of a problem recently encountered when generating dirt samples in SBND (https://github.com/SBNSoftware/sbndcode/issues/410). This pull request implements the missing time shift to resolve the problem.

### DIFF
--- a/src/Tools/Flux/GSimpleNtpFlux.cxx
+++ b/src/Tools/Flux/GSimpleNtpFlux.cxx
@@ -21,6 +21,7 @@
 #include <TSystem.h>
 #include <TStopwatch.h>
 
+#include "Framework/Conventions/Constants.h"
 #include "Framework/Conventions/Units.h"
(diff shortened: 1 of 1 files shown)


nusense (approved):
I think this looks correct, but the proof is in the pudding with an actual study whether this fixes the problem that SBND reported.

sjgardiner:
Lan Nguyen did some high-stats tests of the fix on the FNAL grid, and everything looks good now for SBND. Thanks for having a look, @nusense, I'll go ahead an merge this.

PR #393, opened 2024-03-18
Fix library linking

knoepfel:
These changes are necessary to ensure that ROOT's dictionary loading does not result in a segmentation violation.  The files are adjusted such that the ROOT dictionaries are linked against the necessary libraries.

It would be better to add explicit dependencies in the Make rules.  But this approach at least provides the requisite dependencies such that each GENIE dictionary may be loaded (standalone) by ROOT without creating a segmentation violation.


data/logo/genie_banner_short.txt
nusense, line 1: This need to be left at 999.999.999

VERSION
nusense, line 1: This need to be left at 999.999.999 to indicate that it _isn't_ a tagged version and thus unsanctioned

nusense (changes requested):

nusense (approved):
Version changes have been resolved back to 999.999.999.
The rest looks reasonable, though perhaps not fully thorough, but enough to have the the loading of libraries work again in circumstances that came up at Fermilab

PR #396, opened 2024-03-27
Fix kinematic phase space issue in HybridXSecAlgorithm (due to RosenbluthPXSec)

idkakorin:
Ad hoc solution of problem with inappropriate kinematic phase space reported by @jtenavidal, so she can continue working. 

The reason of problem: it is intended for _LlewelynSmith_,  **but** also used by _Rosenbluth_.

A more thoughtful solution could be:
1. Specify in the configuration file the phase space appropriate for each algorithm.
2. Implement in RosenbluthPXSec the analog of method _LwlynSmithQELCCPXSec::FullDifferentialXSec_.

### DIFF
--- a/src/Framework/EventGen/HybridXSecAlgorithm.cxx
+++ b/src/Framework/EventGen/HybridXSecAlgorithm.cxx
@@ -80,7 +80,17 @@ double HybridXSecAlgorithm::XSec(const Interaction* interaction,
   const XSecAlgorithmI* alg_to_use = this->ChooseXSecAlg( *interaction );
 
   if ( !alg_to_use ) return 0.;
-  else return alg_to_use->XSec( interaction, kps );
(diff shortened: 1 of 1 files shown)


jtenavidal:
After this fix I can get all my events as expected

idkakorin:
After short conversation with @afropapp13 in Slack: return to correct Jacobian for _RosenbluthPXSec_.

afropapp13:
Located documentation from back at the time
https://genie-docdb.pp.rl.ac.uk/DocDB/0002/000214/002/rosenbluth_fix_for_v3.2.pdf

The current version is the correct one so i propose that we close the PR

idkakorin:
Thanks for explanation @afropapp13. However, your suggestion to close PR is premature as it is mainly addressed to fix this issue #397.

jtenavidal (approved):
It now calls correct phase space

PR #398, opened 2024-03-28
Add SuSAv2 EM QE hadron tensors separated by hit nucleon isospin

LiangLiu212:
After last genie Core developer meeting, I made some changes according to the suggestions. 

The existing implementation of the SuSAv2 EM QE model doesn't distinguish between struck protons and neutrons, leading to an incorrect equal cross section for p and n knockout. Liang Liu (@LiangLiu212) has added tensors from Guillermo Megias separated by isospin, together with code to handle them correctly for cross section calculations and event generation.


do_configure.sh
mroda88, line 1: This file to be removed from git

config/SuSAv2QELPXSec.xml
mroda88, line 1: Can you restore the indentation to improve readability?

mroda88 (changes requested):
The tune should disappear, we should use the standard SuSAv2 CMC that are the G21_11*

do_configure.sh
LiangLiu212: This file has been removed.

config/SuSAv2QELPXSec.xml
LiangLiu212: This configuration has been restored from master branch.

PR #401, opened 2024-04-10
Rename KuzminNaumov2016AxialFormFactorModel to MArunAxialFormFactorModel

idkakorin:
The reason for the renaming is to take into account the contributions of all authors. The PR also update references to papers, where the running Ma model is published.


mroda88 (approved):
Looking good also discussed with Costas and Steven Gardiner.

PR #406, opened 2024-04-30
Add a Configuration of matrix in Genie

LiangLiu212:
Since a covariance matrix is needed when I implement the reweight method for Z expansion vector form factor model. Here is a solution to import the covariance matrix from xml configuration.  Covariance matrix can be imported from xml directly or from a plain text file. There are two example:

` <param type="mat-double" name="CCTEST0@CovarianceMatrix" importfile="false" rowdelim=";" coldelim=","> 2.0, 1.0, 2; 1.1, 2.1, 2; 1.1, 2.1, 2; 1.1, 2.1, 3.2 </param>`
   `<param type="mat-double" name="CCTEST1@CovarianceMatrix" importfile="true"> /exp/uboone/app/users/liangliu/Genie/genie/Reweight/data/testmat.dat  </param>`
   
The matrix can be get by [GetParamMat()](https://github.com/LiangLiu212/Generator/blob/f5b1e0aed0fa88b623bb3f6c285f9a9a6dcd2d32/src/Framework/Algorithm/Algorithm.h#L197C1-L203C37)

### DIFF
--- a/data/example/matrix/testmat.dat
+++ b/data/example/matrix/testmat.dat
@@ -1,0 +1,6 @@
+% a matrix with 2 columns and 5 rows
+1.0 1.1
+2.0 2.1
+2.0 2.1
+2.0 2.1
+1.0 5.3
--- a/src/Framework/Algorithm/AlgConfigPool.cxx
+++ b/src/Framework/Algorithm/AlgConfigPool.cxx
@@ -509,7 +509,11 @@ int  AlgConfigPool::AddParameterMatrix  (Registry * r, string pt, string pn, str
   // Each entry will be named pn+"-i"+"-j" where i and j are 
   // index of row and column and replaced by the number
 
-  std::ifstream file(pv);
+  char * GENIE_PATH = std::getenv("GENIE");
+  string filepath = std::string(GENIE_PATH) + "/" + pv;
+  std::cout << filepath << std::endl;
+
+  std::ifstream file(filepath);
   if (!file.is_open()) {
     throw std::runtime_error("Could not open file");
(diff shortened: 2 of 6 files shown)


mroda88 (changes requested):
I love it!! 
But I have 3  requests, all related to the file option. 
 - I would like to see an example of the structure of the file. Not necessarily for me, but because we will forget about that structure. Can you add a file in data? I think you can create a dedicated directory called example, and put a `test_matrix.txt` there. 
  - Can we add a feature to add comments to the matrix file? It does not have to be very complex, but at least if the line starts with `%` we can skip the line. The reason being that if you put a matrix in a file, things will be big and some description will be beneficial. 
  - In the lookup of the file, can you find a way to prioritise looking for the file in $GENIE? This will make the distribution of out file much easier. Something that uses this logic is probably the best https://github.com/GENIE-MC/Generator/blob/d43b421d7f80fa6ef23f5da99ac425fe527cf324/src/Framework/Utils/XmlParserUtils.cxx#L53-L128 
  
I would also have a chat maybe on other possible matrix format you have considered. if you did. Just to avoid picking the first option simply because it was the first one that came to mind.

LiangLiu212:
I add a new method [GetParamMatSym()](https://github.com/LiangLiu212/Generator/blob/8e214bc248b35683e9fa293250c0d344c6e91af9/src/Framework/Algorithm/Algorithm.h#L201-L203) to get the symmetry matrix by using TMatrixTSym.

mroda88 (approved):
Perfect

PR #407, opened 2024-05-14
Update the z expansion vector form factor by using the vec-double

LiangLiu212:
The previous PR is closed. I open a new PR on the top of the current main branch. 

Z expansion vector form factor is recently implemented in genie by Kaushik. The re-weight method is implemented to provide for the study of systematic uncertainty and tune.


sjgardiner (approved):
Looks great, @LiangLiu212, thank you! I rebased the PR on the master branch to make merging even easier. The content is otherwise the same and is approved. I'll merge it now based on the discussion we had this morning with Marco.

PR #408, opened 2024-06-18
Work on Pythia8 interface as a singleton and potential for no Pythia6

nusense:
This is currently a work in progress. Making a pull request to anticipate future pull and allow others to see what is being done.   Supersede pull request #308

PR #410, opened 2024-06-21
changes to XML config files to continue using pythia6 after pythia8 code changes

nusense:
To be used in conjunction with code changes pull request #408


sjgardiner (approved):
Based on core developer meeting discussion today, we will merge the config files designed to preserve Pythia 6 compatibility for now. Eventually we can merge the Pythia-8-only config files from PR #409 when Pythia 6 is officially deprecated. Thanks very much, @nusense!

PR #416, opened 2024-09-16
adding new script to help us address hold jobs which return no output

jtenavidal:
This script is developed by @alons126 and I did some last checks. We need it as for electron jobs at the grid, in which we might run many parallel jobs, some of the parallel jobs might go on hold due to limited resources used by the finished jobs. In this case we end up with less events than expected and we need to re-run the events. This script facilitates the re-running of the events


mroda88 (approved):
LGTM

Issue #82, opened 2020-01-02
Memory leak when loading events from .ghep file


chenel:
There appears to be a memory leak in the deserialization process when events are loaded from a .ghep file.  A very simple sample program will quickly exhaust the memory on any machine with a big enough input file (~1M events seems to do the trick):

```
// CStdLib
#include <iostream>

//ROOT
#include "TFile.h"
#include "TTree.h"

// GENIE
#include "Framework/GHEP/GHepParticle.h"
#include "Framework/Ntuple/NtpMCEventRecord.h"

int main()
{

  TFile gHEPfile("/path/to/large-file.ghep.root", "READ");
  auto gHEPtree = dynamic_cast <TTree *>( gHEPfile.Get("gtree"));

  genie::NtpMCEventRecord *mcrec = nullptr;
  gHEPtree->SetBranchAddress("gmcrec", &mcrec);

  auto nEvts = gHEPtree->GetEntries();
  std::cout << "total number of events:" << nEvts << std::endl;

  // event loop
  gHEPfile.cd();
  for (int iev = 0; iev < nEvts; iev++)
  {
    gHEPtree->GetEntry(iev);
  }

  return 0;
}
```

However, @sjgardiner points out that if the event loop is modified to explicitly delete the event:
```
...
  // event loop
  gHEPfile.cd();
  for (int iev = 0; iev < nEvts; iev++)
  {
    gHEPtree->GetEntry(iev);
    genie::EventRecord* event = mcrec->event;
    delete event;
  }
...
```
then all is well and there's no leak.

I'm not sure if this behavior existed in older GENIE + ROOT combinations but I've observed it in GENIE 3.00.04 + ROOT 6.12.06 and GENIE 3.00.06 + ROOT 6.16.00.

chenel:
Digging a little further, it seems that the problem is likely that ROOT isn't actually calling `NtpMCEventRecord::~NtpMCEventRecord()` when it loads a new entry from the file, but simply writing over it (which _does_ call the constructor for the new entry, and that involves a `new EventRecord` inside `NtpMCEventRecord::Init()`).

steverdennis:
I believe this issue showed up with the switch to ROOT 6. There was an issue open in the old trac system about it but that system seems not to have survived the move away from hepforge.

Assuming it's the same thing, the way I mostly reproduced it was just running a `TTree::Draw` on the GHEP file.

chenel:
[...]

PR #417, opened 2024-09-17
Workaround a memory leak with old ROOT IO

karuboniru:
This is related to a [long existing bug in ROOT](https://github.com/root-project/root/issues/16375), with related workaround in GENIE traced back to [this commit from Apr 2006](https://github.com/GENIE-MC/Generator/commit/64b7b4b759803785202961211342ff1295c98bae). But such workaround by manually calling the `Clean` method after each entry of event being processed is not possible with working with the latest ROOT's RDataFrame analysis framework like I did [in the reweighting testing code](https://github.com/karuboniru/Reweight/blob/professor_reweight/src/Apps/gRwProfTest.cxx).

The bug is that a class is being read from a ROOT file and: 
 - the class contains a pointer to another object
 - the root dict for the class is generated without a "+" in `#pragma link C++ class ClassName+;`.

Since the bug is related to the old IO implementation in ROOT they are not going to fix it. But there are 2 approaches to fix or workaround this:
 1. move to new IO implementation, always specify "+" in `LinkDef.h`
 2. Use user defined `Streamer` to free the memory before the pointer is overwritten by TBuffer.

The first one seemd to be feasible, but the fatal issue is that the old and new IO don't compatible. The root file generated with new IO method won't be recognized by old dictionaty and vice versa. Developers from ROOT [did provide an example](https://github.com/karuboniru/root_memory_leak_report/pull/1/files) to show how to partially workaround the incompatibility with the version number tricks to keep old files readable, but it would still break the compatibility from the other side.

So this PR is the conservative approach, the second one, the IO and file format is kept the same. This should not be breaking anything, the only difficulty it would introduce is when we change the defination of `NtpMCEventRecord` we would have to update the definition of `NtpMCEventRecord::Streamer` here accordingly. (But at that time we can really move to the new IO from ROOT since change the definition of a class will introduce incompatibility)

### DIFF
--- a/src/Framework/Ntuple/LinkDef.h
+++ b/src/Framework/Ntuple/LinkDef.h
@@ -12,7 +12,7 @@
 #pragma link C++ class genie::NtpMCJobConfig;
 #pragma link C++ class genie::NtpMCRecHeader;
 #pragma link C++ class genie::NtpMCRecordI;
-#pragma link C++ class genie::NtpMCEventRecord;
+#pragma link C++ class genie::NtpMCEventRecord-;
 #pragma link C++ class genie::NtpWriter;
 
 #endif
--- a/src/Framework/Ntuple/NtpMCEventRecord.cxx
+++ b/src/Framework/Ntuple/NtpMCEventRecord.cxx
@@ -10,6 +10,7 @@
 
 #include "Framework/EventGen/EventRecord.h"
 #include "Framework/Messenger/Messenger.h"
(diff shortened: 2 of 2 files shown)


karuboniru:
refs: https://github.com/GENIE-MC/Generator/issues/82

nusense (approved):
These changes look to do what they're supposed to do and shouldn't conflict with our use of the Clean() method, so I'll approve them

PR #418, opened 2024-10-04
Fix some obvious memory leak

karuboniru:
This is a side-product when invegesting some memory leak with GENIE in JUNO (which turns out to be issues in certain ROOT version). But profiling of memory allocation and deallocation (`fsanitize=leak`) inside GENIE showed some memroy leak. I cleaned some of them (that happens multiple times).

Most issues are kind of equivalent to 
```cpp
int p = *(new int); // but the new is hidden in function call that not that obvious. 
```
Also, I add a function `TrimSpacesClean` next to `TrimSpaces` to clean memory allocated by `xmlNodeListGetString`.


mroda88 (approved):
LGTM.
Shall we merge or do you want to add more stuff?

karuboniru:
If you think its good I think we can merge.

PR #421, opened 2024-11-08
bring back the "correlated nucleon tail"

sjgardiner:
The correlated tail from previous work disappeared with the most recent modifications to the LFG model here.  Forcing the computed removal energy to be nonnegative makes sense for momenta drawn from the bulk of the LFG distribution; negative removal energy ordinarily means the nucleon is not bound (i.e. outside the nucleus).  However, for momenta drawn from the correlated tail, they will *always* be negative relative to the calculation from the LFG model: their associated KE is much larger because their *internal* (pairwise) potential energy offsets it.  Obviously the LFG computation can't be aware of that.

Here we take cases where the LFG calculation winds up negative and simply reassign them a (more) sensible removal energy.  The removal energy is sampled from a Maxwell-Boltzmann distribution, as is done for the "correlated spectral function" in spectral function fits.  Here we use rough parameters taken from estimates by Artur Ankowski given in private communication to J. Wolcott (Ankowski paper forthcoming).  The values thrown by the MB distribution are capped such that the off-shell struck nucleon in the initial state has nonnegative mass, which otherwise causes NaNs when trying to boost into its rest frame.

Because the default behavior of the "AR23" CMC would otherwise be changing under the feet of folks who want to use it, this PR also includes some changes to the AR23 configuration so that it should preserve the previous behavior.  Here we also include a "N24" (NOvA 2024) CMC that turns it on.


LiangLiu212 (approved):
Approved

PR #422, opened 2024-11-19
properly protect against use of pythia8 object when only building pythia6

nusense:
add some cases of "#ifdef __GENIE_PYTHIA8_ENABLED__ ... #endif"


sjgardiner (approved):
Looks good, @nusense, thanks for the quick fix!

PR #423, opened 2024-11-20
disable pythia8

jtenavidal:
This pull request changes the setup scripts for GENIE so we disable pythia 8 until we have access in the cvmfs area


mroda88 (approved):
LGTM

PR #427, opened 2024-12-20
Add Liang Liu to long banner and README

sjgardiner:
We had not yet updated the collaboration list to include Liang Liu (postdoc at Fermilab working with Steven G). This PR includes his name in both places where the authors appear in the repository.


LiangLiu212 (approved):
Approved

PR #428, opened 2025-01-10
Make the use of Professor2 fully optional 

nusense:
Primarily change GOPT_WITH_PROFESSOR2 to GOPT_ENABLE_PROFESSOR2 for consistency sake and to keep Make.config_no_paths contain the option for use on the Reweight side.  Set the #define in GBuild.h  accordingly.  Also leave the PROFESSOR2_{INCLUDES|LIBRARIES} Makefile variables blank if it's not enabled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,7 @@ else
 endif
 
 lib-professor2-build: FORCE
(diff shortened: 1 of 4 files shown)


karuboniru (approved):

mroda88 (approved):
This looks good to me

PR #429, opened 2025-01-10
Jan 10 2025: adding config dirs for G24_20j,k,l, sanity check that Q2…

afropapp13:
… is positive


sjgardiner (approved):
Looks good, thanks @afropapp13. Approved.

PR #430, opened 2025-01-14
Changes to XML files to support use of Pythia8 rather than Pythia6

nusense:
this overrides a previous, now deleted (#409) pull request that wasn't sufficient.  This also updates some Messenger files.


sjgardiner (approved):
Looks good, @nusense, thank you! Are any changes needed here in light of your other Pythia-related changes on PRs targeting v3.8?

nusense:
This PR should be orthogonal to any other Pythia related changes in that it only targets XML model selection and messenger config files.  It should not be in conflict with CommonDecay.xml which is in #483.

PR #431, opened 2025-01-15
nsl library is no longer required

nusense:
Nobody knows why this `-lnsl` library was linked in, but it doesn't seem necessary any more and isn't available by default on AlmaLinux9 nodes.


mroda88 (approved):
LGTM

PR #435, opened 2025-02-11
Implement of NucleusGenerator

LiangLiu212:
Implement of NucleusGenerator to replace the old FermiMover and VertexGenerator.

NucleusGenerator is inherited from EventRecordVisitorI. This interface facilitates the use of various implementations of nuclear models.
Now I create two implementations: NucleusGenINCL and NucleusGenTraditional.

NucleusGenINCL nuclear model is used for both initial state and FSI model to have a consistent physics results.

NucleusGenTraditional combines the FermiMover and VertexGenerator.

### DIFF
--- a/src/Physics/NuclearState/NucleusGenINCL.cxx
+++ b/src/Physics/NuclearState/NucleusGenINCL.cxx
@@ -1,0 +1,336 @@
+//____________________________________________________________________________
+/*!
+
+\class    genie::NucleusGenINCL
+
+\brief    It visits the event record & computes a Fermi motion momentum for
+          initial state nucleons bound in nuclei.
+          Is a concrete implementation of the EventRecordVisitorI interface.
+
+\author   Liang Liu <liangliu \at fnal.gov>
+          Fermi National Accelerator Laboratory
+
+\created  October 17, 2024
+
+\cpright  Copyright (c) 2003-2024, The GENIE Collaboration
+          For the full text of the license visit http://copyright.genie-mc.org
+          
+*/
+//____________________________________________________________________________
+
+#include "Framework/Conventions/GBuild.h"
+#ifdef __GENIE_INCL_ENABLED__
+#include <cstdlib>
+
+#include <TLorentzVector.h>
+#include <TVector3.h>
+#include <TParticlePDG.h>
+#include <TMath.h>
+
+#include "Framework/Algorithm/AlgFactory.h"
+#include "Framework/Algorithm/AlgConfigPool.h"
+#include "Framework/Conventions/Constants.h"
+#include "Framework/Conventions/Units.h"
+#include "Physics/NuclearState/NucleusGenINCL.h"
+
+#include "Physics/NuclearState/NuclearModel.h"
+#include "Physics/NuclearState/NuclearModelI.h"
+#include "Framework/EventGen/EVGThreadException.h"
+#include "Framework/GHEP/GHepRecord.h"
+#include "Framework/GHEP/GHepParticle.h"
+#include "Framework/GHEP/GHepStatus.h"
+#include "Framework/GHEP/GHepFlags.h"
+#include "Framework/Interaction/Interaction.h"
+#include "Framework/Messenger/Messenger.h"
+#include "Physics/NuclearState/FermiMomentumTablePool.h"
+#include "Physics/NuclearState/FermiMomentumTable.h"
+#include "Framework/Numerical/RandomGen.h"
+#include "Framework/ParticleData/PDGLibrary.h"
+#include "Framework/ParticleData/PDGUtils.h"
+#include "Framework/ParticleData/PDGCodes.h"
+#include "Framework/Utils/KineUtils.h"
+#include "Physics/NuclearState/NuclearUtils.h"
+#include "Physics/NuclearState/INCLNucleus.h"
+
+using namespace genie;
+using namespace genie::constants;
+
+//___________________________________________________________________________
+NucleusGenINCL::NucleusGenINCL() :
+EventRecordVisitorI("genie::NucleusGenINCL")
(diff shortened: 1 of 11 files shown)


src/Physics/NuclearState/NucleusGenINCL.cxx
```diff
@@ -0,0 +1,336 @@
...
+  Target *       tgt         = init_state  -> TgtPtr();
+
+  // do nothing for non-nuclear targets
+  if(!tgt->IsNucleus()) return;
+
+  TLorentzVector * p4 = tgt->HitNucP4Ptr();
+
+  // do nothing if the struct nucleon 4-momentum was set (eg as part of the
+  // initial state selection)
+  if(p4->Px()>0 || p4->Py()>0 || p4->Pz()>0) return;
```
nusense, line 186: what if all three are non-zero but negative?  Perhaps `p4->Vect()->Mag2() > 0` trading three multiplies and one comparison, against 3 `TMath::Abs()` and three comparisons?

src/Physics/NuclearState/NucleusGenINCL.cxx
```diff
@@ -0,0 +1,336 @@
...
+
+  // access the hit nucleon and target nucleus at the GHEP record
+  GHepParticle * nucleon = evrec->HitNucleon();
+  GHepParticle * nucleus = evrec->TargetNucleus();
+  assert(nucleon);
```
nusense, line 191: Just in case you didn't know.  the `<cstdlib>` `assert()` is typically disabled in non-debug code.   So, they're good for testing with code compiled with debug options, but ineffective for non-debug built code.  Lots of people don't know that.

src/Physics/NuclearState/NucleusGenINCL.cxx
```diff
@@ -0,0 +1,336 @@
...
+  assert(nucleus);
+
+  // initialize INCL nucleus model
+  // INCL nucleus model sample all nucleons with r-p correlation
+  INCLNucleus *incl_nucleus = INCLNucleus::Instance();
+  G4INCL::Nucleus *incl_nuc = incl_nucleus->getNuclues();
+  TLorentzVector p4tgt;
```
nusense, line 198: MIght be useful here to explicitly call out, in a comment, the units used by INCL and GENIE to explain the use of `1000`.

src/Physics/NuclearState/NucleusGenINCL.cxx
```diff
@@ -0,0 +1,336 @@
...
+  double   hit_nucleon_energy = incl_nucleus->getHitNucleonEnergy();
+  double   w  = incl_nucleus->getRemovalEnergy();
+  //-- update the struck nucleon 4p at the interaction summary and at
+  // the GHEP record
+  p4->SetPx(p3.Px()/1000.);
+  p4->SetPy(p3.Py()/1000.);
+  p4->SetPz(p3.Pz()/1000.);
+  p4->SetE (hit_nucleon_energy/1000.);
+  LOG("NucleusGenINCL", pINFO) << "Momentum";
+  p4->Print();
```
nusense, line 219: Again consider `PrintUtils` `Vec3AsString()` so that can be part of the LOG line and controlled by the message service.

src/Physics/NuclearState/NucleusGenINCL.cxx
```diff
@@ -0,0 +1,336 @@
...
+  }//daughters
+
+  TParticlePDG * remn = 0;
+  int ipdgc = pdg::IonPdgCode(A, Z);
+  remn = PDGLibrary::Instance()->Find(ipdgc);
+  if(!remn) {
+    LOG("HadronicVtx", pFATAL)
+          << "No particle with [A = " << A << ", Z = " << Z
+                            << ", pdgc = " << ipdgc << "] in PDGLibrary!";
+    assert(remn);
```
nusense, line 296: The earlier comment here about `assert` being a no-op for non-debug code makes this section not do what I think you want it to.   Even if you're hoping that the branch never gets triggered it's best that it does what it is intended to do.

src/Physics/NuclearState/NucleusGenTraditional.cxx
```diff
@@ -0,0 +1,111 @@
...
+//____________________________________________________________________________
+void NucleusGenTraditional::LoadConfig(void)
+{
+
+
+  fFermiMover = nullptr;
+  fVertexGenerator = nullptr;
+  fFermiMover = dynamic_cast<const EventRecordVisitorI *> (this->SubAlg("FermiMover"));
+  fVertexGenerator = dynamic_cast<const EventRecordVisitorI *> (this->SubAlg("VertexGenerator"));
+
```
nusense, line 108: Just dropping a marker here, because I'm not sure of how SubAlg's work, that it should be checked that `fFermiMover` and `fVertexGenerator` get `Configured()` if there are non-default cases.

src/Physics/NuclearState/NucleusGenerator.cxx
```diff
@@ -0,0 +1,87 @@
...
+  this->LoadConfig();
+}
+//____________________________________________________________________________
+void NucleusGenerator::LoadConfig(void)
+{
+  RgKey nuclkey = "NucleusGenerator";
+  fNucleusGen = nullptr;
+  //fNuclModel = dynamic_cast<const NucleusGenerator *> (this->SubAlg(nuclkey));
+  //
+  LOG("NucleusGenerator", pINFO) << "Hello world!";
```
nusense, line 83: :-)

nusense (commented):
There's lots of good stuff here.  Don't let the sheer number of comments I made discourage you; many are just attempts to make the code better and not criticisms of what you're trying to achieve.

PR #442, opened 2025-04-04
Now searches for root by looking for root-config

luketpickering:
rather than deprecated ROOTSYS Env Var


nusense (approved):
This looks like a sensible change
Makefile depends on using `root-config` so it has to be there to succeed.

PR #445, opened 2025-04-22
Constrained lower limit for EDnu integration.

AlessandroRuggeri:
Update to the DarkNeutrino module cross sections.
Applied a lower cut to the range of integration over E_DNu in order to correct the unphysical behaviour of the total cross-sections.
The cut corresponds to a Q^2(Max) < 1 GeV^2, above which the differential cross-sections are negligible.


mroda88 (approved):
Thanks @AlessandroRuggeri . This is fixes the splines.

mroda88:
[GENIE_range_correction_Fe56_check_12042025.pdf](https://github.com/user-attachments/files/19974606/GENIE_range_correction_Fe56_check_12042025.pdf)
 This is the proof of the new splines which are in agreement with other generators.

PR #446, opened 2025-05-04
Add handling of the macOS arm64 architecture to Make.include

sjgardiner:
`root-config --arch` returns `macosxarm64` on my new Fermilab-issued MacBook Pro, and this confuses the GENIE build system. Adding this as a possibility to `Make.include` builds fine with otherwise normal `macosx64` settings.


nusense (approved):
This looks fine to me

PR #448, opened 2025-06-11
Properly Sample Lorentz Invariant Phase Space in Pion Production FSI

gputnam:
The current pion production FSI model in GENIE intranuke hA samples the angle of the outgoing nucleon (theta3), the angle of the target nucleon (theta4), and the center-of-momentum outgoing nucleon energy (E3CM) uniformly.

This does not properly sample the lorentz invariant phase space. As is the case in general for spherical coordinates, the proper quantity to sample is cos(theta), not theta. Furthermore, to ensure Lorentz Invariance it is necessary to weight the sampled point by a product of the outgoing momenta (see https://pdg.lbl.gov/2021/reviews/rpp2021-rev-kinematics.pdf eq. 49.20).

I have also added the possibility to bias the outgoing momenta by exp(B*t), where B is a constant and t is the invariant mass of the 4-momentum transfer from the probe nucleon to the target-pion system. This feature enables the kinematics to be set as they are in INCL, which uses a bias of B = 6 GeV^-2 (see https://doi.org/10.1016/j.nuclphysa.2011.07.004 section 2.2). The bias is set to 0 by default, to maintain the default GENIE behavior.

### DIFF
--- a/config/HAIntranuke2018.xml
+++ b/config/HAIntranuke2018.xml
@@ -52,7 +52,7 @@ DelRNucleon         double  Yes   mult. factor for nucleon de-Broglie wavelength
     <param type="bool"   name="INUKE-DoCompoundNucleus"> true  </param>
     <param type="bool"   name="INUKE-XsecNNCorr">        true  </param>
 
- 
+
 
     <param type="double" name="HAINUKE-DelRPion">        0.0   </param>
     <param type="double" name="HAINUKE-DelRNucleon">     0.0   </param>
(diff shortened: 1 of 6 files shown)


sjgardiner:
Thanks for catching this, @gputnam, much appreciated. I have a proposed change to the implementation to avoid relying on a hard-coded value for the 3-body phase space bias parameter. See my related PR on your fork for details. Otherwise, this looks good to me.

sjgardiner (dismissed):
With the updates to allow configuration and avoid hard-coding, this looks good.

sjgardiner:
Looks good, but I still need to update the hA2025 source code (already merged) to incorporate this fix there as well. I'll update the PR and re-approve once I'm done with that.

PR #452, opened 2025-06-28
correctly record the phase space used in RES calculator

chenel:
After some adjustments to the variables acceptance/rejection sampling for the RES calculator was using a couple years ago, when RES events were stored, the phase space variables used were not correctly stored in the event---instead the wrong set of variables were.  This PR fixes that oversight.

### DIFF
--- a/src/Physics/Resonance/EventGen/RESKinematicsGenerator.cxx
+++ b/src/Physics/Resonance/EventGen/RESKinematicsGenerator.cxx
@@ -196,8 +196,12 @@ void RESKinematicsGenerator::ProcessEventRecord(GHepRecord * evrec) const
         double M = init_state.Tgt().HitNucP4().M();
         kinematics::WQ2toXY(E,M,gW,gQ2,gx,gy);
 
-        // set the cross section for the selected kinematics
-        evrec->SetDiffXSec(xsec,acceptKps);
+        // set the cross section for the selected kinematics.
+        // note that we're saving to the evt record in the more familiar "W*Q2" space
+        // rather than the "W*Q2D" (precomputed dipole) space that's used above
+        // for generation efficiency in the accept-reject loop
+        double J = kinematics::Jacobian(interaction, kPSWQD2fE, kPSWQ2fE);
+        evrec->SetDiffXSec(J * xsec, kPSWQ2fE);
 
         // for uniform kinematics, compute an event weight as
         // wght = (phase space volume)*(differential xsec)/(event total xsec)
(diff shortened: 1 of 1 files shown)


idkakorin:
Hi Jeremy,

Since Q2D is more of an auxiliary variable, while users are more familiar with Q2, I think it would be better to modify line 199 as follows:
```
double J = kinematics::Jacobian(interaction, kPSWQD2fE, kPSWQ2fE);
evrec->SetDiffXSec(J * xsec, kPSWQ2fE);
```
**and leave all other lines unchanged.**

Just to clarify the reasoning: the maximum for the acceptance-rejection method is found in the kPSWQD2fE phase space, and the current xsec is evaluated in the same space — which is correct. However, when storing the cross section value into the event record, it would make more sense to transform it into the more familiar kPSWQ2fE space, using the Jacobian.

This way, the internal logic of the sampling remains untouched, while the stored value is more intuitive for downstream users.

idkakorin (changes requested):
Hi Jeremy,

Since Q2D is more of an auxiliary variable, while users are more familiar with Q2, I think it would be better to modify line 199 as follows:

```
double J = kinematics::Jacobian(interaction, kPSWQD2fE, kPSWQ2fE);
evrec->SetDiffXSec(J * xsec, kPSWQ2fE);
```
**and leave all other lines unchanged.**

Just to clarify the reasoning: the maximum for the acceptance-rejection method is found in the kPSWQD2fE phase space, and the current xsec in the method is evaluated in the same space — which is correct. However, when storing the cross section value into the event record, it would make more sense to transform it into the more familiar kPSWQ2fE space, using the Jacobian.

This way, the internal logic of the sampling remains untouched, while the stored value is more intuitive for downstream users.

chenel:
Oh!  Except @idkakorin I think if I left everything else unchanged, there would still be an error.  Line 209-211 unchanged had 
```
          double vol     = kinematics::PhaseSpaceVolume(interaction,kPSWQ2fE);
          double totxsec = evrec->XSec();
          double wght    = (vol/totxsec)*xsec;
 ```
But I think since `xsec`  was already computed using `WQ2D` kinematics, I think this weight is incorrect (it would be multiplying a phase space volume for `WQ2` by a xsec computed for `WQ2D`).  So my counterproposal is to leave the other lines as I've changed them (including 209, which currently reads `double vol     = kinematics::PhaseSpaceVolume(interaction,acceptKps);`) but accept your proposal for 199-200.  See my amended commit above.

chenel:
Yes, but I'm talking about `xsec`, not `totalxsec`, in the calculation.  Isn't multiplying `vol` by `xsec` in line 201 self-inconsistent, since they were computed using different phase spaces?

chenel:
> double J = kinematics::Jacobian(interaction, kPSWQD2fE, kPSWQ2fE);
> xsec *= J;
> evrec->SetDiffXSec(xsec, kPSWQ2fE);

Okay, that would also achieve the same thing as what I did.

idkakorin:
Agreed — that makes sense.

idkakorin (approved):
That’s fine by me.

PR #454, opened 2025-07-01
Remove use of 'hash' to check for root-config in the configure script.

sjgardiner:
Using root-config directly is more portable.


LiangLiu212 (approved):
Approve

PR #455, opened 2025-07-02
fix distclean of Apps; had wrong source name and wildcarding pattern

nusense:
distclean for the Apps was always wonky
turns out it was cleaning out $(GENIE_BIN_PATH) instead of $(GENIE_BIN_INSTALLATION_PATH) due to incorrect use of make's patsubst and the use of $(GENIE_BIN_DIR) where it should have been $(GENIE_BIN_PATH)


sjgardiner (approved):
Looks great @nusense, thanks for fixing this!

PR #462, opened 2025-09-22
Move GSystUncertaintyTable XML configuration to Reweight

sjgardiner:
The GSystUncertaintyTable algorithm lives in the Reweight repository, but the XML configuration lives here in Generator. The mismatch is due to the historical lack of a config/ area in Reweight. However, since a config/ area now exists, it makes much more sense to move the GSystUncertaintyTable XML file there.

This PR should be merged at the same time as the corresponding one in Reweight.


nusense (approved):
This looks fine in conjunction with Reweight pull 43

Issue #438, opened 2025-03-04
Incorrect NC scaling in QPMDISPXSec


ts4051:
In `QPMDISPXSec.cxx`, the EM scaling variable (`fEMScale`) is applied both to EM and weak-NC events. And the weak-NC scaling variable  (`fNCScale`) is not used. Presumably this is a bug.

The issue can be seen here: https://github.com/GENIE-MC/Generator/blob/master/src/Physics/DeepInelastic/XSection/QPMDISPXSec.cxx#L175C44-L175C52

jtenavidal:
Hi, Yes it is a bug, 
This will be fixed once the changes for DIS are merged in. 
See the fix here: https://github.com/GENIE-MC/Generator/blob/0c3dbeefa756b8fba0d93bb09fe3303073e6a187/src/Physics/DeepInelastic/XSection/QPMDISPXSec.cxx#L249

Júlia

PR #465, opened 2025-10-20
Bodek Yang Model Update

jtenavidal:


jtenavidal:
Validation of current branch against master - Running Default model configuration with both branches to guarantee we can reproduce old results: 

For EM (e+p)
image

jtenavidal:
I made few changes in the current version regarding the nuclear models. These shouldn't affect ongoing validations at the free nucleon level. 

The first correction is to allow the BY00DISNuclearModel to have configurable parameters. 
I also modified the AxialSea and AxialValance K factor calculations to remove a hardcoded value which was equivalent to MA2/4. MA comes from the configuration and therefore should be configurable to avoid issues if the user uses a different elasitc MA. 

Next, I will change the current logic of how nuclear corrections are applied. In the original code, the correction (f) was applied directly to the structure functions, F3 and F2, and subsequently also in F1 and F5. However, this correction factor should be seen as a modification of the partonic pdfs. I changed the code so that these are implemented when computing the PDFs. 

This will facilitate the implementation of more fancy nuclear corrections, which affect differently valance and sea quarks. The nuclear model will remain configurable by the user.

jtenavidal:
I noticed an issue while adding the merge request. The QPMDMDISStrucFuncBase.cxx, which as I understand is specific to DM, needs to also change accordingly. Need to revisit this issue separately. 

For instance, it uses the same K factors and nuclear modifications as BY.

AndersenFred:
#438

PR #466, opened 2025-11-16
Improve APFEL detection: support modern CMake builds (no .la)

idkakorin:
Configure: accept APFEL libraries without .la (support .so and .a)

Replaced legacy check for libAPFEL.la with a broader test that accepts
libAPFEL.so, libAPFEL.a, or libAPFEL.la. This allows using modern APFEL
installations built with CMake (which no longer generate .la files).


nusense (approved):
These changes seem sensible and make the configure more flexible in finding an appropriate APFEL library.

PR #468, opened 2025-12-10
Adds a `SplinePostProcessor` singleton class to Generator

kjplows:
For the cross-section algorithms that rely on `NewQELXSec` to integrate, MC throws of the struck nucleon are required to compute the cross-section. This leads to a "jitter" in the graph of the cross-section as a function of neutrino energy.

This PR adds a new singleton class `genie::SplinePostProcessor`, that acts after spline knots have been collected in `XSecSplineList`, and applies a [Gaussian kernel smoother](https://en.wikipedia.org/wiki/Kernel_smoother#Gaussian_kernel_smoother) to the knots, "smoothing" the spline.

The `SplinePostProcessor` acts only on algorithms that it has been configured to run on (one can add these as keys in the `SplinePostProcessor.xml` file), and only if it has been switched on at the configuration. By default, the `SplinePostProcessor` is configured to act on `NievesQELCCPXSec/ZExp`.

One may specify a sequence of "stops", or reference energies, at which the width of the window is given by a corresponding entry in a sequence of "widths". In between "stops", the width is interpolated linearly.
One may also specify additional "stops" and "widths" per neutrino PDG; this is done by specifying the additional string `@Pdg=<probe PDG>` in the configuration key. In this XML file, a separate key is set for tau neutrinos, which have significantly higher thresholds than muon or electron neutrinos.

Some validation plots, in the form of "Brazil bands" follow: first, for 12C, 16O, and 40Ar, and for each of the (anti)neutrino flavours, one hundred runs of `gmkspl -e 10 -n 100` (using tune `AR23_20i_00_000`) were generated; for each of the spline xmls made, for each of the energies sampled, the 100 results for σ were collected into a distribution.

A band of (<σ> - 2υ, <σ> + 2υ) is coloured in yellow; a band of (<σ> - υ, <σ> + υ) is coloured in green, where <σ>, υ are the sample mean and variance of this distribution.
Then, a spline is generated, with the `SplinePostProcessor` switched off: this is plotted as red points. The jitter is evident, with a fraction of points lying outside the "2 sigma" yellow band.
Finally, a spline with the same seed is generated, with the `SplinePostProcessor` switched on, and plotted as the dashed blue line. The graph of the cross-section is both less "jittery", and now lies within the band, indicating a better agreement with the "expected" cross-section.
Shown here: numu on 40Ar, nue on 12C, nutau on 16O.

40Ar_numu
12C_nue
16O_nutau

### DIFF
--- a/config/Messenger.xml
+++ b/config/Messenger.xml
@@ -205,6 +205,7 @@
   <priority msgstream="SpectralFunc1">         NOTICE </priority>
   <priority msgstream="SpectralFunc">          NOTICE </priority>
   <priority msgstream="Spline">                INFO   </priority>
+  <priority msgstream="SplinePostProcessor">   WARN   </priority>
   <priority msgstream="Target">                INFO   </priority>
   <priority msgstream="Utils">                 INFO   </priority>
   <priority msgstream="Vtx">                   WARN   </priority>
--- a/config/SplinePostProcessor.xml
+++ b/config/SplinePostProcessor.xml
@@ -25,8 +25,8 @@ Widths                    vec-double  no    Reference widths of the Gaussian ker
     <param type="vec-double" delim="," name="Stops"> 0.1,0.2,0.5,1.0,2.0,5.0,10.0,50.0,100.0 </param>
     <param type="vec-double" delim="," name="Widths"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
     <!-- separate list for nutau due to high thresholds -->
-    <param type="vec-double" delim="," name="Stops@Pdg=16,-16"> 2.8,3.0,3.5,4.0,5.0,7.5,10.0,50.0,100.0 </param>
-    <param type="vec-double" delim="," name="Widths@Pdg=16,-16"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+    <param type="vec-double" delim="," name="Stops@Pdg=16,-16"> 2.8,3.0,4.0,5.0,10.0 </param>
+    <param type="vec-double" delim="," name="Widths@Pdg=16,-16"> 0.005,0.01,0.03,0.075,0.1 </param>
 
     <param type="bool" name="UsePostProcessing"> true </param>
 
--- a/config/SplinePostProcessor.xml
+++ b/config/SplinePostProcessor.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+
+<!--
+Configuration for the SplinePostProcessor
+Tool that forces Gaussian kernel smoothing on selected algorithms
+
+Algorithm Configurable Parameters:
+..............................................................................................................................
+Name                      Type        Opt   Comment                                           Default
+..............................................................................................................................
+
+************                                *******************************
+set Default:                                Generic configuration parameter
+************                                *******************************
+HandledAlgorithms         vec-string  no    Keys of algorithms handled by this tool.
+Stops                     vec-double  no    Reference neutrino energies at which the Gaussian kernel has given width
+Widths                    vec-double  no    Reference widths of the Gaussian kernel at specific neutrino energies
+-->
+
+<alg_conf>
+
+  <param_set name="Default">
+
+    <param type="vec-string" delim="," name="HandledAlgorithms"> genie::NievesQELCCPXSec/ZExp </param>
+    <param type="vec-double" delim="," name="Stops"> 0.1,0.2,0.5,1.0,2.0,5.0,10.0,50.0,100.0 </param>
+    <param type="vec-double" delim="," name="Widths"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+
+  </param_set>
+
+</alg_conf>
--- a/config/SplinePostProcessor.xml
+++ b/config/SplinePostProcessor.xml
@@ -24,6 +24,10 @@ Widths                    vec-double  no    Reference widths of the Gaussian ker
     <param type="vec-string" delim="," name="HandledAlgorithms"> genie::NievesQELCCPXSec/ZExp </param>
     <param type="vec-double" delim="," name="Stops"> 0.1,0.2,0.5,1.0,2.0,5.0,10.0,50.0,100.0 </param>
     <param type="vec-double" delim="," name="Widths"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+    <!-- separate list for nutau due to high thresholds -->
+    <param type="vec-double" delim="," name="Stops@Pdg=16,-16"> 2.8,3.0,3.5,4.0,5.0,7.5,10.0,50.0,100.0 </param>
+    <param type="vec-double" delim="," name="Widths@Pdg=16,-16"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+
     <param type="bool" name="UsePostProcessing"> true </param>
 
   </param_set>
--- a/config/SplinePostProcessor.xml
+++ b/config/SplinePostProcessor.xml
@@ -24,6 +24,7 @@ Widths                    vec-double  no    Reference widths of the Gaussian ker
     <param type="vec-string" delim="," name="HandledAlgorithms"> genie::NievesQELCCPXSec/ZExp </param>
     <param type="vec-double" delim="," name="Stops"> 0.1,0.2,0.5,1.0,2.0,5.0,10.0,50.0,100.0 </param>
     <param type="vec-double" delim="," name="Widths"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+    <param type="bool" name="UsePostProcessing"> true </param>
 
   </param_set>
 
--- a/src/Apps/gMakeSplines.cxx
+++ b/src/Apps/gMakeSplines.cxx
@@ -196,7 +196,6 @@ int main(int argc, char ** argv)
       GEVGDriver driver;
       driver.SetEventGeneratorList(RunOpt::Instance()->EventGeneratorList());
       driver.Configure(init_state);
-      LOG("gmkspl", pDEBUG) << "Configured for init_state:" << init_state;
       driver.CreateSplines(gOptNKnots, gOptMaxE);
     }
   }
@@ -374,8 +373,7 @@ PDGCodeList * GetNeutrinoCodes(void)
   vector<string>::const_iterator iter;
   for(iter = nuvec.begin(); iter != nuvec.end(); ++iter) {
     list->push_back( atoi(iter->c_str()) );
-  }
-  
+  }  
   return list;
 }
 //____________________________________________________________________________
--- a/src/Apps/gMakeSplines.cxx
+++ b/src/Apps/gMakeSplines.cxx
@@ -196,6 +196,7 @@ int main(int argc, char ** argv)
       GEVGDriver driver;
       driver.SetEventGeneratorList(RunOpt::Instance()->EventGeneratorList());
       driver.Configure(init_state);
+      LOG("gmkspl", pDEBUG) << "Configured for init_state:" << init_state;
       driver.CreateSplines(gOptNKnots, gOptMaxE);
     }
   }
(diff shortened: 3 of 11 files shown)


config/Messenger.xml
```diff
@@ -205,6 +205,7 @@
   <priority msgstream="SpectralFunc1">         NOTICE </priority>
   <priority msgstream="SpectralFunc">          NOTICE </priority>
   <priority msgstream="Spline">                INFO   </priority>
+  <priority msgstream="SplinePostProcessor">   WARN   </priority>
   <priority msgstream="Target">                INFO   </priority>
```
nusense, line 209: I suggest `SplinePostProcessor` entries in the other `Messenger_XYZZY.xml` files as well; have them consistent with the general theme of the file.

src/Apps/gMakeSplines.cxx
```diff
@@ -373,7 +373,7 @@ PDGCodeList * GetNeutrinoCodes(void)
   vector<string>::const_iterator iter;
   for(iter = nuvec.begin(); iter != nuvec.end(); ++iter) {
     list->push_back( atoi(iter->c_str()) );
-  }
+  }
```
nusense, line 376: Unnecessary addition of trailing blank space.

src/Framework/Numerical/SplinePostProcessor.cxx
```diff
@@ -0,0 +1,333 @@
+//____________________________________________________________________________
+/*
+ Copyright (c) 2003-2025, The GENIE Collaboration
```
nusense, line 3: Since this is going into the code base in 2026, perhaps that should be the copyright date (here and elsewhere)

config/SplinePostProcessor.xml
```diff
@@ -0,0 +1,35 @@
...
+
+  <param_set name="Default">
+
+    <param type="vec-string" delim="," name="HandledAlgorithms"> genie::NievesQELCCPXSec/ZExp </param>
+    <param type="vec-double" delim="," name="Stops"> 0.1,0.2,0.5,1.0,2.0,5.0,10.0,50.0,100.0 </param>
+    <param type="vec-double" delim="," name="Widths"> 0.005,0.01,0.025,0.05,0.1,0.25,0.47,2.34,4.7 </param>
+    <!-- separate list for nutau due to high thresholds -->
+    <param type="vec-double" delim="," name="Stops@Pdg=16,-16"> 2.8,3.0,4.0,5.0,10.0 </param>
+    <param type="vec-double" delim="," name="Widths@Pdg=16,-16"> 0.005,0.01,0.03,0.075,0.1 </param>
+
```
nusense, line 30: Does the stops/widths list for nu_tau(_bar) need to go as far as the standard one?    Why or why not?
Also, what happens above the upper stop?  We typically generate splines out to E=1000 GeV, though out that high there tends not to be much jitter.

src/Framework/Numerical/SplinePostProcessor.h
```diff
@@ -0,0 +1,136 @@
...
+
+  class Spline;
+
+  class SplinePostProcessor: public SplinePostProcessorI {
+
+  public:
+    SplinePostProcessor();
+    SplinePostProcessor(std::string name);
+    SplinePostProcessor(std::string name, std::string config);
+    ~SplinePostProcessor();
```
nusense, line 80: Usually for singletons the constructor and destructor are private and the only public way to get the one-and-only instance is via Instance().   This prevents users from accidentally creating one outside the singleton framework.   I don't know if we have any other singletons that are algorithms, and haven't reviewed all the code to see how this works in actuality here so I don't know whether it is possible for this code.   This is something to consider.
Also a common pattern in singletons is a Cleaner to destroy the one instance when the program is shutting down.   If you look at some of the other singletons you'll see code (here for AlgFactory in the private portion of the header) like:
<pre>
  //! singleton cleaner
  struct Cleaner {
      void DummyMethodAndSilentCompiler() { }
      ~Cleaner() {
         if (AlgFactory::fInstance !=0) {
            delete AlgFactory::fInstance;
            AlgFactory::fInstance = 0;
         }
      }
  };
  friend struct Cleaner;
</pre>
 and in the `.cxx` file:
<pre>
AlgFactory * AlgFactory::Instance()
{
  if(fInstance == 0) {
    static AlgFactory::Cleaner cleaner;
    cleaner.DummyMethodAndSilentCompiler();

    fInstance = new AlgFactory;
  }
  return fInstance;
}
</pre>
Then when the static is unloaded, it cleans up the associated instance.

src/Framework/Numerical/SplinePostProcessor.cxx
```diff
@@ -0,0 +1,333 @@
...
+  this->GetParamVect( "Stops", fStops );
+  this->GetParamVect( "Widths", fWidths );
+  fStopMap[0]  = fStops;
+  fWidthMap[0] = fWidths;
+
+  // If there are any more keys, typically delineated with "@Pdg=...", add these to the map
+  // Similar architecture to NuclearModelMap's implementation
+  RgIMap entries = (this->GetConfig()).GetItemMap();
+  for(RgIMap::const_iterator it = entries.begin(); it != entries.end(); ++it){
+    const std::string  key  = it->first; // Notice this prepends "N" and appends "s"
```
nusense, line 125: Huh, is it understood why the key gets prepended by `N` and have a `s` appended?  I'd hate to have this block of code depend on something that was accidental or (compiler) implementation dependent.

src/Framework/Numerical/SplinePostProcessor.cxx
```diff
@@ -0,0 +1,333 @@
...
+	  int pdg = std::stoi(subs.substr(0, pos));
+	  LOG("SplinePostProcessor", pDEBUG) << "Adding key " << pdg << " to stops";
+	  subs.erase(0, pos+1); // "," has length 1
+	  std::vector<double> stops; stops.resize(n);
+	  for( int i = 0; i < n; ++i ) {
+	    RgKey tmp_key = Algorithm::BuildParamVectKey(comm_name, i);
+	    this->GetParam( tmp_key, stops[i] );
+	  }
+	  fStopMap[pdg] = stops;
+	} // split strings
```
nusense, line 156: This block of code for `if( key.find("Stops")...` (and the latter one for `Widths`) have a very wonky indentation, in the GitHub representation, that doesn't follow the code flow.   The line following the `if` is 2 spaces to the left from the `if`.   It might be that the files are using tabs (which we discourage because their interpretation can wildly vary) instead of spaces.

Also, just to note, that given `subs` (without having appended a `,` on `dup_name`) one could also have had the elements extracted using
<pre>
 vector&lt;string&gt; pdgstrvec = genie::utils::str::Split(subs,",");
 while ( size_t indx=0, indx &lt; pdgstrvec.size(); indx++) {
    int pdf = std::stoi(pdgstrvec[indx]);
...
 }
</pre>
I'm not sure I'm advocating changing what's here, but more making you aware of some GENIE utility functions.

src/Framework/Numerical/SplinePostProcessor.cxx
nusense: Looking further, I'm guessing the key with the prepended `N` and appended `s` is just the "count" of the elements of the key without those.  So the "Notice this ..." is more like "We're looking for the keys with..."

src/Framework/Numerical/SplinePostProcessor.cxx
```diff
@@ -0,0 +1,333 @@
...
+    if(E[i] > *it_stop && E[i] < stops.back()) { // Update iterators
+      it_stop++; it_width++;
+      w1 = w2; w2 = *(it_width);
+    }
+    
+    double rat = activation(*(it_stop-1), *(it_stop));
+    double width = w1 + (w2 - w1) * rat; // between w1 and w2
+    double num = 0.0; double den = 0.0;
+    for(int j = 0; j < NKnots; j++) {
+      double w = std::exp(-0.5 * std::pow((E[i] - E[j]) / width, 2.0)); // Gauss
```
nusense, line 277: This line is the heart of the Gaussian kernel smoother, and probably needs a comment to that effect ahead of the line rather than a simple `// Gauss`. :-)

src/Framework/Numerical/SplinePostProcessor.cxx
```diff
@@ -0,0 +1,333 @@
...
+
+  // If the xmin and/or xmax are outside first / last stop, add some with copies of width
+  if( xmin < stops.front() ) {
+    stops.insert(stops.begin(), xmin);
+    widths.insert(widths.begin(), widths.front());
+  }
+  if( xmax > stops.back() ) {
+    stops.emplace_back(xmax);
+    widths.emplace_back(widths.back());
+  }
```
nusense, line 324: Ah, I think this answers an earlier question about what happen to when the stops/widths don't cover the range of the spline.  I read this as the first (or last) entry in the stops/widths are duplicated at the front or back to fully cover the range.

src/Framework/Utils/XSecSplineList.cxx
```diff
@@ -267,6 +267,24 @@ void XSecSplineList::CreateSpline(const XSecAlgorithmI * alg,
 
   }
 
+  fPostProcessor = *( SplinePostProcessor::Instance() );
```
nusense, line 270: This and line 103 of the header.  Because `fPostProcessor` is not a pointer, what happens here is a full copy of the one given by `SplinePostProcessor::Instance()`.   I think what you should be doing is making `fPostProcessor` a pointer, assigning the pointer from the `Instance()` just for convenience and converting the following uses from `fPostProcessor.X()` to `fPostProcessor->X()`.   Otherwise you're making unnecessary copies of a fairly hefty data structure.

src/Framework/Utils/XSecSplineList.h
```diff
@@ -97,6 +99,8 @@ class XSecSplineList {
   virtual ~XSecSplineList();
 
   static XSecSplineList * fInstance;
+  
+  SplinePostProcessor fPostProcessor;
```
nusense, line 103: See comment in `XSecSplineList.cxx` about this being an actual object and not a pointer.

nusense (changes requested):
A first pass review

kjplows:
Hi @nusense , thanks for the detailed comments! I'll start implementing these as soon as I can.
Cheers!

config/SplinePostProcessor.xml
kjplows: Yeah, I definitely didn't give too much thought to the stops / widths for nutau, nutaubar. But as you noted -- above the upper stop, the last width is used. Now that I think about it, perhaps the best solution is to add a stop at 1 TeV with some width. 

(The scaling between points is linear, currently; line 262 and friends of `SplinePostProcessor.cxx`. So if last 2 stops are e.g. at 10 GeV, 1e3 GeV; at E =100 GeV, the activation becomes (100-10)  / (1000 - 10) = 1/11 , meaning the width at that point is width(10) + (width(1e3) - width(10)) / 11 . Probably should add an option for logarithmic scaling, come to think of it.)

src/Framework/Numerical/SplinePostProcessor.h
kjplows: I think that makes sense, if turning the `Instance` into a pointer means I can still assign an `fPostProcessor` at `XSecSplineList` level then happy to do this!

src/Framework/Numerical/SplinePostProcessor.cxx
kjplows: Yeah this caused me some consternation. I think this is caused by `AlgConfigPool` for parameter vectors (actually, while looking for this code I also found that it does a similar thing for matrices with `Nrow`, `Ncol`!). Pasting comment in `AlgConfigPool::AddParameterVector` (`AlgConfigPool.cxx::L409`) here:

```
 // Adds a configuration parameter vector
  // It is simply add a number of entries in the Registy
  // The name scheme starts from the name and it goes like
  // 'N'+pn+'s' that will be an integer with the number of entries.
  // Each entry will be named pn+"-i" where i is replaced by the number
```

src/Framework/Numerical/SplinePostProcessor.cxx
kjplows: Huh, I never knew about tabs vs spaces could resolve in wonky indentation! Thanks for the catch, I'll make changes. 

I'll give the utility a go. It looks like a more elegant solution.

sjgardiner:
As noted in the discussion of PR #495, this PR is superseded. I will close it now to avoid confusion.

PR #472, opened 2026-01-21
Implementation of 1p1h and npnh Martini et al channels

lavinia-russo:
This fork implements the 1p1h and npnh channels of the Martini et al. model. 
The chosen tune is G24_12a. All details can be found in the PRD article (DOI: 10.1103/v8hq-kppp). 

A presentation of this implementation was given at the GENIE Core Meeting on Monday, June 16th. 

As requested during the meeting, the old non-working npnh implementation from 2016 has been removed.


data/evgen/hadron_tensors/martini/README.md
nusense, line 1: This is very nice README file. I long for more like this.

src/Physics/HadronTensors/MartiniMECHadronTensorModel.h
```diff
@@ -0,0 +1,46 @@
...
+\brief    Creates hadron tensor objects for calculations of MEC
+          cross sections using the Martini approach
+
+\author   Steven Gardiner <gardiner \at fnal.gov>
+          Fermi National Accelerator Laboratory
+
+\created  April 26, 2019
+
+\cpright  Copyright (c) 2003-2023, The GENIE Collaboration
+          For the full text of the license visit http://copyright.genie-mc.org
```
nusense, line 15: Bump the copyright year (here and presumably other places)

src/Physics/HadronTensors/TabulatedLabFrameHadronTensor.cxx
```diff
@@ -374,7 +374,42 @@ double genie::TabulatedLabFrameHadronTensor::dSigma_dT_dCosTheta_rosenbluth(
   double Tl      = interaction->Kine().GetKV(kKVTl);
   double cos_l   = interaction->Kine().GetKV(kKVctl);
   double ml      = interaction->FSPrimLepton()->Mass();
-
+   
+  /*
+  double xsec_1 = dSigma_dT_dCosTheta_rosenbluth(14, 0.5, m_probe, 329.3416/1000., 0.7391376742, ml, 0);
+  std::cout << "Enu is 0.5 and Tl is 329.3416 cos = 0.7391376742 and xsec is " << xsec_1 << std::endl;
+  std::cout << "\n\n";
+
```
nusense: I'd turn these `std::cout` into Messenger service calls.   If they're going to a new stream don't forget to put entries in the `Messenger_XYZZY.xml` files.

Oh, I see this whole block is `/* */` commented out.  I see it's making some calculations for special cases and then printing the result.   Is this in order to compare with some external numbers?   If so, perhaps those should be also entered here as reference.   Otherwise, is there ever a reason to uncomment the block?

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
```diff
@@ -106,6 +107,19 @@ void MECGenerator::ProcessEventRecord(GHepRecord * event) const
...
   }
+  else if (fXSecModel->Id().Name() == "genie::MartiniEricsonChanfrayMarteauMECPXSec2024") {
+      event->Print();
+      this -> SelectMartiniLeptonKinematics(event);
+      event->Print();
+      this -> AddTargetRemnant(event);
+      event->Print();
+      this -> GenerateNSVInitialHadrons(event);
+      event->Print();
+      // Note: this method in `MECTensor/MECTensorGenerator.cxx` appeared to be a straight
```
nusense, line 110: I question the multitude of of `event->Print()` calls here.  I presume that's debugging code and should be stripped out.

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
```diff
@@ -1125,6 +1139,266 @@ void MECGenerator::SelectSuSALeptonKinematics(GHepRecord* event) const
...
+  Interaction* interaction = event->Summary();
+  Kinematics* kinematics = interaction->KinePtr();
+
+  // Choose the appropriate minimum Q^2 value based on the interaction
+  // mode (this is important for EM interactions since the differential
+  // cross section blows up as Q^2 --> 0)
+  double Q2min = genie::controls::kMinQ2Limit; // CC/NC limit
+  if ( interaction->ProcInfo().IsEM() ) Q2min = genie::utils::kinematics
+    ::electromagnetic::kMinQ2Limit; // EM limit
+
```
nusense, line 1146: For readability, I'd suggest moving the whole "then" part of this `if` to the line after the conditional.

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
```diff
@@ -1125,6 +1139,266 @@ void MECGenerator::SelectSuSALeptonKinematics(GHepRecord* event) const
...
+        double percent_deviation = 200. * ( XSec - XSecMax ) / ( XSecMax + XSec );
+
+        if ( percent_deviation > fSuSAMaxXSecDiffTolerance ) {
+          LOG( "Kinematics", pFATAL ) << "xsec: (curr) = " << XSec
+            << " > (max) = " << XSecMax << "\n for " << *interaction;
+          LOG( "Kinematics", pFATAL )
+             << "*** Exceeding estimated maximum differential cross section";
+          std::terminate();
+        }
```
nusense, line 1266: Is there any way to recover here rather than invalidating the file by an unclean exit?

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
```diff
@@ -1125,6 +1139,266 @@ void MECGenerator::SelectSuSALeptonKinematics(GHepRecord* event) const
...
+	if ( interaction->ProcInfo().IsEM() ) {
+      // calculate ppFraction in the EM case
+	  ppFraction = dynamic_cast< const SuSAv2MECPXSec* >( fXSecModel )
+          ->PairRatio( interaction ,"ppFraction");
+
+      LOG("MEC", pINFO) << "Test for pp: "
+                        << "; xsec = " << XSec << "; pp_fraction = " << ppFraction
+                        << "; random number val = " << myrand_pp;
+	}
+
```
nusense, line 1303: This block of `IsEM()` code has an indentation inconsistent with the surrounding code.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+    else {
+      MAXLOG("MartiniEricsonChanfrayMarteauMEC", pWARN, 10)
+          << "Asked to scale to a nucleus "
+          << target_pdg << " which we don't know yet.";
+      return 0;
+    }
+  }
+
+  // The MartiniEricsonChanfrayMarteau-MEC hadron tensors are defined using the same conventions
+  // as the Valencia MEC model, so we can use the same sort of tensor
```
nusense, line 138: I just realized the MartiniEricsonChanfrayMarteau just need to drop Marteau, so this can be MEC-MEC.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  // *** Enforce the global Q^2 cut (important for EM scattering) ***
+  // Choose the appropriate minimum Q^2 value based on the interaction
+  // mode (this is important for EM interactions since the differential
+  // cross section blows up as Q^2 --> 0)
+  double Q2min = genie::controls::kMinQ2Limit; // CC/NC limit
+  if ( interaction->ProcInfo().IsEM() ) Q2min = genie::utils::kinematics
+    ::electromagnetic::kMinQ2Limit; // EM limit
+
```
nusense, line 183: Again, suggest moving the split qualified value to just the 2nd line.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+  // Compute the cross section using the hadron tensor
+  double xsec = tensor->dSigma_dT_dCosTheta_rosenbluth(interaction, Delta_Q_value);
+
+  // This scaling should be okay-ish for the total xsec, but it misses
+  // the energy shift. To get this we should really just build releveant
+  // hadron tensors but there may be some ways to approximate it.
+  // For more details see Guille's thesis: https://idus.us.es/xmlui/handle/11441/74826
+  if ( need_to_scale ) {
+    FermiMomentumTablePool * kftp = FermiMomentumTablePool::Instance();
```
nusense, line 203: Best I can determine, for this routine, `need_to_scale` is *always* false when it gets here.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  // Apply given overall scaling factor
+  xsec *= fXSecScale;
+
+  // Scale given a scaling algorithm:
+  if( fMECScaleAlg ) xsec *= fMECScaleAlg->GetScaling( * interaction ) ;
+
+  if ( kps != kPSTlctl ) {
+    LOG("MartiniEricsonChanfrayMarteauMEC", pWARN)
```
nusense: Might make sense to move this to an early exit before doing all the work.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+  // binding energy in the Valencia model but this effect is already
+  // in Guille's tensors so its set it to 0.
+  // However, additional corrections may be necessary:
+  double Delta_Q_value = Qvalue( * interaction ) ;
+
+  // Compute the cross section using the hadron tensor
+  double xsec_all = tensor->dSigma_dT_dCosTheta_rosenbluth(interaction, Delta_Q_value);
+
+  double ratio;
+
```
nusense, line 316: strictly speaking if the passed in `final_state_ratio` isn't `"pnFraction"` or `"ppFraction"` then this method is going to return an undefined value for ratio.  Good practice would be to initialize ration to 0.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
```diff
@@ -0,0 +1,517 @@
...
+    if( !fQvalueShifter ) {
+      good_config = false ;
+      LOG("MartiniEricsonChanfrayMarteauMECPXSec2024", pERROR) << "The required QvalueShifterAlg does not exist. AlgId is : " << SubAlg("QvalueShifterAlg")->Id() ;
+    }
+  }
+
+  if( ! good_config ) {
+    LOG("MartiniEricsonChanfrayMarteauMECPXSec2024", pERROR) << "Configuration has failed.";
+    exit(78) ;
+  }
```
nusense, line 527: If you're going to exit() here, perhaps the preceding LOG should be `pFATAL`

src/Physics/QuasiElastic/EventGen/QELEventGeneratorMartini.cxx
```diff
@@ -0,0 +1,668 @@
...
+  Interaction * interaction = event->Summary();
+  Kinematics * kinematics = interaction->KinePtr();
+
+  // Choose the appropriate minimum Q^2 value based on the interaction
+  // mode (this is important for EM interactions since the differential
+  // cross section blows up as Q^2 --> 0)
+  double Q2min = genie::controls::kMinQ2Limit; // CC/NC limit
+  if ( interaction->ProcInfo().IsEM() ) Q2min = genie::utils::kinematics
+    ::electromagnetic::kMinQ2Limit; // EM limit
+
```
nusense, line 104: Ugh.  Hate to harp on it but these splits just bug me.

src/Physics/QuasiElastic/EventGen/QELEventGeneratorMartini.cxx
```diff
@@ -0,0 +1,668 @@
...
+    TParticlePDG * remn = 0;
+    int ipdgc = pdg::IonPdgCode(A, Z);
+    remn = PDGLibrary::Instance()->Find(ipdgc);
+    if(!remn) {
+        LOG("HadronicVtx", pFATAL)
+            << "No particle with [A = " << A << ", Z = " << Z
+            << ", pdgc = " << ipdgc << "] in PDGLibrary!";
+        assert(remn);
+    }
```
nusense, line 335: Also, is there some less catastrophic approach to this, such as throwing an exception and dropping the event?

src/Physics/QuasiElastic/XSection/MartiniQELPXSec.cxx
```diff
@@ -0,0 +1,342 @@
...
+  // Apply given overall scaling factor
+  xsec *= fXSecScale;
+
+  if ( kps != kPSTlctl ) {
+    LOG("MartiniQE", pWARN)
+      << "Doesn't support transformation from "
+      << KinePhaseSpace::AsString(kPSTlctl) << " to "
+      << KinePhaseSpace::AsString(kps);
+    xsec = 0.;
+  }
```
nusense: Again, I think this test could be at the beginning of the method.

src/Physics/QuasiElastic/XSection/MartiniQELPXSec.cxx
```diff
@@ -0,0 +1,342 @@
...
+  bool isnu  = pdg::IsNeutrino(nu);
+  bool isnub = pdg::IsAntiNeutrino(nu);
+  bool is_chgl = pdg::IsChargedLepton(nu);
+
+  bool prcok = ( proc_info.IsWeakCC() && ((isP && isnub) || (isN && isnu)) )
+    || ( proc_info.IsEM() && is_chgl && (isP || isN) );
+  if ( !prcok ) return false;
+
+  return true;
+}
```
nusense, line 273: These two lines of `return` could be condensed down to `return prcok`, no?

nusense (commented):
Overall very clean written and organized code.  I picked a few nits but generally I have little issue with what's there.

sjgardiner:
Adding link to the DocDB presentation for the GENIE core meeting talk from last June: https://genie-docdb.pp.rl.ac.uk/cgi-bin/private/ShowDocument?docid=535.

src/Physics/HadronTensors/TabulatedLabFrameHadronTensor.cxx
lavinia-russo: the whole block can actually be removed. it was a test during the first attempts of the implementation

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
lavinia-russo: I guess so... I copied it from the Susav2 case (L97-104 of the same file), so maybe it's worth strip that out as well. @sjdolan you confirm?

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
lavinia-russo: not sure... maybe @sjdolan can help

src/Physics/QuasiElastic/EventGen/QELEventGeneratorMartini.cxx
lavinia-russo: totally right!

src/Physics/HadronTensors/MartiniMECHadronTensorModel.h
lavinia-russo: yes you're right! should I update the copyright myself?

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
lavinia-russo: you're right, I am gonna check with Stephen D if it is what we wanted (and so I would remove the if block)... need to work on that

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
lavinia-russo: I took it from SuSav2MECPXSec.cxx file in the same directory... maybe it makes sense to change it there as well? @sjdolan what do you think?

src/Physics/QuasiElastic/EventGen/QELEventGeneratorMartini.cxx
lavinia-russo: not sure, it comes from QELEventGeneratorSuSA.cxx as well :/

src/Physics/HadronTensors/MartiniMECHadronTensorModel.h
nusense: Yes, please.  Here and everywhere.

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
sjdolan: Yes, this is just left over from debugging! These lines can be removed (from here and the SuSA case)

src/Physics/QuasiElastic/EventGen/QELEventGeneratorMartini.cxx
sjdolan: I think think it makes sense that the user should be made very very aware if they're somehow passing an A and Z which gives an ion that isn't in the PDG library. 

I think the MEC generator behaviour is just not to check ipdgc at all, which probably isn't better.

It probably makes sense for the behaviour to be unified across GENIE, but I also don't think we want to just drop the check here. Perhaps this is something to add for a future general consideration of how GENIE behaves if passed an IonPdgCode the PDG doesn't know about.

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
lavinia-russo: ok I added the `if(tensor_pdg != target_pdg) need_to_scale = true;` line!

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
sjdolan: Once we find one xsec over the max then technically we know our acceptance probability for all other events was incorrect. I think if this ever happens with more than very very rare frequency it needs some expert attention. 

I guess if it happens once in something like 1E6 events it's probably okay, but the fix then would be to implement a counter for bad events which I think would get quite messy. Since we never his this in any of the event samples we generated (which covered the full phase space of 2p2h events on all the targets we implemented) I think this probably isn't worth giving too much attention to (but we're of course open to discussion about it!).

src/Physics/Multinucleon/XSection/MartiniEricsonChanfrayMarteauMECPXSec2024.cxx
sjdolan: I agree! @lavinia-russo can you move it for SuSA as well?

src/Physics/Multinucleon/EventGen/MECGenerator.cxx
nusense: If it's so rare, might it be worth considering just complaining and either  accepting it (but knowing that it's under-represented) or potentially aborting the _event_ (so the chosen kinematics are even more underrepresented) but in either case *not* terminating?  I'm always leery of abort/terminate which implies that things are such a mess that one can't continue and the whole file to date gets invalidated because it's not a clean end of processing.

nusense: If. you're worried that one might be letting this pass too frequently perhaps a local static counter and terminating if it exceeds some threshold

nusense: Given that my proposed idea _was_ considered (and rejected) and that this failure mode is very rare, I'm just going to "resolve" this thread.

nusense (dismissed):
Essentially all my queries/suggestions have been resolved, so I'll approve, but not merge, this.

nusense:
Sorry the correct user is not `lars-p` but @larsb-p

nusense (approved):
The differences since the previous review look sensible.  I'll approve it.

sjgardiner:
Closing in favor of the rebased (easier to merge) and slightly more cleaned-up PR #507.

PR #475, opened 2026-03-12
address Roberts review comments

idkakorin:
Applied the changes discussed with Robert for the **DCC_Sato** branch. They are not branch-specific and are therefore relevant for the **master** branch as well.


nusense (approved):
These were essentially reviewed in a separate PR and were addressed there.  Approving.

PR #476, opened 2026-03-16
Fix GCC 14 warnings (deprecated-copy, unused variables, sign-compare,shadow)

idkakorin:
This PR performs a general cleanup of compiler warnings observed when building GENIE with newer compilers (notably GCC ≥14). While investigating some unrelated changes locally, I noticed that the build output contains a fairly large number of warnings originating from legacy patterns in the codebase (e.g., deprecated copy semantics, unused variables/parameters, sign–compare issues, shadowed variables, etc.). None of these affect runtime behavior, but they do make the build output unnecessarily noisy and can obscure real issues.

The goal of this PR is therefore purely maintenance-oriented: to reduce warning noise and remove a few small pieces of unused or redundant code. The changes are mechanical and intentionally minimal in scope.

Summary of the cleanup:

* Addressed `-Wdeprecated-copy` warnings by explicitly defining defaulted assignment operators in a few small utility classes (e.g., `Range*`, `AlgId`, `RgAlg`).
* Removed or simplified unused variables and parameters that triggered `-Wunused-*` warnings.
* Fixed several `-Wsign-compare` and `-Wshadow` cases.
* Replaced a few outdated constructs with equivalent modern C++ forms where this could be done safely.
* Removed a small number of unused helper functions that were no longer referenced anywhere in the code.
* Minor stylistic adjustments (e.g., eliminating unnecessary temporary objects).

These modifications should not affect physics results or runtime behavior; they only touch compiler diagnostics and small structural details. One practical side effect is that compilation output becomes much cleaner and slightly faster to process, since the compiler no longer emits large volumes of warnings.

I tested the changes locally with GCC 14 and verified that the GENIE build proceeds cleanly and produces the same binaries as before (aside from the expected warning reductions).

If possible, it would be great to get a quick review so that this cleanup could be included in the upcoming **3.8 release**, helping keep the build output clean for users compiling with modern toolchains.

Thanks!

### DIFF
--- a/src/Apps/gMakePhotonStrucFunc.cxx
+++ b/src/Apps/gMakePhotonStrucFunc.cxx
@@ -68,9 +68,8 @@ extern "C" void externalsetapfellept_(double* x, double* q, int* irep, double* x
 
 
 // Requires a global cache of pdf_cache LHAPDF::PDF
-void externalsetapfellept_(double* x, double* q, int* irep, double* xl, double* xf){
+void externalsetapfellept_(double* x, double* q, /*int* irep,*/ double* xl, double* xf){
   
-  (void)irep;
   if (*x >= 1 || *x < 0) {
     for ( int i=0; i<13; i++ ) xf[i] = 0.;
     for ( int i=0; i<7; i++ )  xl[i] = 0.;
--- a/src/Apps/gMakePhotonStrucFunc.cxx
+++ b/src/Apps/gMakePhotonStrucFunc.cxx
@@ -69,6 +69,8 @@ extern "C" void externalsetapfellept_(double* x, double* q, int* irep, double* x
 
 // Requires a global cache of pdf_cache LHAPDF::PDF
 void externalsetapfellept_(double* x, double* q, int* irep, double* xl, double* xf){
+  
+  (void)irep;
   if (*x >= 1 || *x < 0) {
     for ( int i=0; i<13; i++ ) xf[i] = 0.;
     for ( int i=0; i<7; i++ )  xl[i] = 0.;
@@ -108,7 +110,7 @@ void externalsetapfellept_(double* x, double* q, int* irep, double* xl, double* 
 #endif
 
 //____________________________________________________________________________
-int main(int argc, char ** argv)
+int main(void)
 {
 
   string basedir = "";
--- a/src/Framework/Numerical/Interpolator2D.cxx
+++ b/src/Framework/Numerical/Interpolator2D.cxx
@@ -163,30 +163,40 @@ double Interpolator2D::Eval(const double & x, const double & y) const
 double Interpolator2D::DerivX(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
 }
 //____________________________________________________________________________
 double Interpolator2D::DerivY(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
 }
 //____________________________________________________________________________
 double Interpolator2D::DerivXX(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
 }
 //____________________________________________________________________________
 double Interpolator2D::DerivXY(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
 }
 //____________________________________________________________________________
 double Interpolator2D::DerivYY(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
 }
 //____________________________________________________________________________
--- a/src/Physics/BoostedDarkMatter/EventGen/DMDISInteractionListGenerator.cxx
+++ b/src/Physics/BoostedDarkMatter/EventGen/DMDISInteractionListGenerator.cxx
@@ -93,7 +93,7 @@ InteractionList * DMDISInteractionListGenerator::CreateInteractionList(
       if(fSetHitQuark) {
         // Add interactions for all possible hit (valence or sea) quarks
 
-        multimap<int,bool> hq = this->GetHitQuarks(interaction);
+        multimap<int,bool> hq = this->GetHitQuarks();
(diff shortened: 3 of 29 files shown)


src/Apps/gMakePhotonStrucFunc.cxx
```diff
@@ -69,6 +69,8 @@ extern "C" void externalsetapfellept_(double* x, double* q, int* irep, double* x
 
 // Requires a global cache of pdf_cache LHAPDF::PDF
 void externalsetapfellept_(double* x, double* q, int* irep, double* xl, double* xf){
+  
+  (void)irep;
```
nusense: I presume this is to avoid a case of `unused variable`.   Traditionally we've handled this by commenting out the name in the function args, Ala
<pre>
void externalsetapfellept_(double* x, double* q, int* /* irep */, double* xl, double* xf){
</pre>

src/Framework/Numerical/Interpolator2D.cxx
```diff
@@ -163,30 +163,40 @@ double Interpolator2D::Eval(const double & x, const double & y) const
 double Interpolator2D::DerivX(const double & x, const double & y) const
 {
   assert(!"Method requires GSL version 2 or higher.");
+  (void)x;
+  (void)y;
   return -999;
```
nusense, line 168: Again, I think the preferred approach to unused arguments is to avoid naming them, i.e.
<pre>
double Interpolator2D::DerivX(const double & /* x */, const double & /* y */) const
</pre>
In this file, here and 4 times further down.

src/Physics/BoostedDarkMatter/EventGen/DMDISInteractionListGenerator.cxx
```diff
@@ -151,10 +151,10 @@ multimap<int,bool> DMDISInteractionListGenerator::GetHitQuarks(
 {
 // Set (PDG code, from-sea flag) for all possible hit quarks for the input
 // interaction
-
+  (void)interaction;
```
nusense: another instance of using `(void)X` rather than commenting out the naming of the arg.

src/Physics/Common/NormGenerator.cxx
```diff
@@ -33,7 +33,7 @@ EventRecordVisitorI("genie::NormGenerator")
 NormGenerator::NormGenerator(string config):
 EventRecordVisitorI("genie::NormGenerator")
 {
-
+  (void)config;
```
nusense, line 36: Still not a fan of the `(void)X` paradigm  :-)

src/Physics/Decay/Pythia8Decayer2023.cxx
```diff
@@ -234,6 +234,8 @@ bool Pythia8Decayer2023::Decay(int decay_particle_id, GHepRecord * event) const
 
   return true;
 #else
+  (void) decay_particle_id;
+  (void) event;
```
nusense, line 238: This use of `(void)X` is probably necessary due to the `#if / #else / #endif` code unless one puts the function signature itself in a `#if / #else / #endif` block, which isn't pretty.

src/Physics/HadronTensors/TabulatedHadronTensorModelI.cxx
```diff
@@ -31,6 +31,7 @@ namespace {
   /// Converts a string to a genie::HadronTensorType_t value. If the string
   /// does not correspond to a valid tensor type, kHT_Undefined
   /// is returned, and the ok flag is set to false
+  /*
```
nusense: Is it this block of code is never used?   Perhaps add a note to that effect when commenting it out?

src/Physics/HELepton/XSection/PhotonCOHPXSec.cxx
```diff
@@ -109,13 +109,13 @@ double PhotonCOHPXSec::XSec(
   return xsec;
 }
 //____________________________________________________________________________
-double PhotonCOHPXSec::F2_Q(double Q, double r0) const
+double PhotonCOHPXSec::F2_Q(double Q, double r) const
```
nusense: Is there a reason to change this arg name from `r0` to `r`?  The cited paper seems to use `r_0` (i.e. `r` sub `0`) in formula A.3.

nusense (approved):
A few comments that might be addressed but I'm going to approve this.

src/Physics/HadronTensors/TabulatedHadronTensorModelI.cxx
idkakorin: Yes, it appears that this block is never used anywhere in the codebase. I removed it during the warning cleanup since it was triggering an unused-function warning.

Good point about documenting it — I’ll add a short note in the comment to make that clearer.

src/Apps/gMakePhotonStrucFunc.cxx
idkakorin: Good point — I switched this one to the commented-argument style as you suggested.

src/Framework/Numerical/Interpolator2D.cxx
idkakorin: Here I kept the `(void)x` / `(void)y` pattern to avoid changing the interface defined in the header. Since this PR is mainly a warning-cleanup, I tried to keep the function signatures unchanged where possible.

src/Physics/BoostedDarkMatter/EventGen/DMDISInteractionListGenerator.cxx
idkakorin: Since this helper method is private and does not override any interface method, I removed the unused argument from the signature rather than keeping a commented parameter.

src/Physics/Common/NormGenerator.cxx
idkakorin: Here I kept the (void) pattern to avoid changing the interface defined in the header.

src/Physics/Decay/Pythia8Decayer2023.cxx
idkakorin: Yes, exactly. Because of the `#if / #else` block the arguments end up unused in one branch, so `(void)x` seemed cleaner than wrapping the whole function signature in preprocessor directives.

src/Physics/HELepton/XSection/PhotonCOHPXSec.cxx
idkakorin: Good catch — the rename was originally just to avoid a `-Wshadow` warning due to the global `r0` defined earlier in the file. To keep the notation closer to the paper, I changed the argument name to `R0` instead.

PR #477, opened 2026-03-16
comment out unused parameter names instead of using (void)

idkakorin:


nusense (approved):
Okay, these seem sensible.  Commenting out the named variable in a function argument list will avoid the "unused variable" warning without disturbing the arg list.

PR #479, opened 2026-03-20
fix memory leaks; add memory sanitizer

idkakorin:
### Summary

This PR adds support for AddressSanitizer (ASan) in the GENIE build system and fixes several memory-related issues uncovered during testing.

### Changes

* Added a new `--enable-asan` configure option
* Propagated sanitizer flags (`-fsanitize=address -fno-omit-frame-pointer`) to both compilation and linking stages
* Updated multiple Makefile targets to ensure sanitizer flags are consistently applied when building shared libraries
* Fixed several memory issues (including leaks) identified with ASan, including leaks in `TuneId`

### Notes

* ASan support is intended for debugging and development builds
* Leak detection may be noisy due to external dependencies (e.g. ROOT), so it can be optionally disabled via `ASAN_OPTIONS=detect_leaks=0`
* No impact on default builds unless explicitly enabled

### Motivation

The changes improve debugging capabilities and help identify memory safety issues (e.g. use-after-free, leaks) that are otherwise difficult to trace with standard tools. In particular, enabling ASan allowed detection of memory leaks in `TuneId`.


src/Framework/Utils/TuneId.cxx
```diff
@@ -207,8 +207,12 @@ bool TuneId::CheckDirectory() {
...
   std::vector<std::string> paths = utils::str::Split(pathlist,":;,");
 
-  string top_path = gSystem->ExpandPathName( paths[0].c_str() ) ;
-  string def_path = gSystem->ExpandPathName( utils::xml::GetXMLDefaultPath().c_str() ) ;
+  char* tmp = gSystem->ExpandPathName(paths[0].c_str());
+  std::string top_path = tmp;
+  delete[] tmp;
+  tmp = gSystem->ExpandPathName( utils::xml::GetXMLDefaultPath().c_str() ) ;
+  string def_path = tmp;
+  delete[] tmp;
```
nusense, line 215: Good catch.   I really hate that `gSystem->ExpandPathName()` returns a pointer to a char string that needs to be managed; it's so awkward to use.

src/make/Make.include
```diff
@@ -610,6 +610,12 @@ CXXFLAGS := $(ENV_CXXFLAGS) $(CXXFLAGS)
 CPPFLAGS  = $(CPP_INCLUDES)
 CCFLAGS   = $(ROOT_DICT_GEN_INCLUDES)
 CFLAGS    = -g
+# --- AddressSanitizer ---
+ifeq ($(strip $(GOPT_ENABLE_ASAN)),YES)
+  CXXFLAGS += $(GOPT_WITH_SANITIZER_FLAG)
+  LDFLAGS  += $(GOPT_WITH_SANITIZER_FLAG)
+  LIBRARIES += $(GOPT_WITH_SANITIZER_FLAG)
+endif
```
nusense, line 618: I'm not sure I understand the addition to the `LIBRARIES` list.   Isn't the flag only necessary to be added to `LDFLAGS` and the `LIBRARIES` variable just the list of actual libraries with no flags?

nusense (approved):

src/make/Make.include
idkakorin: You’re absolutely right — thanks for pointing that out.

I’ve removed the sanitizer flag from `LIBRARIES`, since it should only contain actual libraries. Instead, I slightly adjusted how shared objects are built by propagating the sanitizer flags through `SOCMD`, so they are consistently applied at link time.

This keeps the behavior correct without relying on `LIBRARIES`, while still covering all the cases where `LDFLAGS` is not used in the current build setup.

PR #480, opened 2026-03-25
Fix missing Single Pion Production handling in gspl2root

idkakorin:
This PR fixes an issue where Single Pion Production (SPP) processes were not handled in `gSplineXml2Root.cxx`.

The SPP process type was introduced relatively recently and was not accounted for in `gspl2root`, leading to incomplete conversion of spline data.

This change adds proper support for SPP processes, which is important for analyses and models relying on Single Pion Production.


nusense (approved):
This all seems sensible for my understanding of how 1pion is now handled

PR #481, opened 2026-03-30
fix potential uninitialized variable warnings

idkakorin:
Fix uninitialized variable warnings (no physics changes)

### DIFF
--- a/src/Physics/QuasiElastic/EventGen/QELEventGeneratorSM.cxx
+++ b/src/Physics/QuasiElastic/EventGen/QELEventGeneratorSM.cxx
@@ -458,6 +458,7 @@ double QELEventGeneratorSM::ComputeMaxXSec(const Interaction * interaction) cons
     double xsec_max = -1;
     double tmp_xsec_max = -1;
     double Q20, v0;
+    bool initialized = false;
     const int N_Q2 = 32;
     const InitialState & init_state = interaction -> InitState();
     Target * tgt = init_state.TgtPtr();
@@ -486,11 +487,12 @@ double QELEventGeneratorSM::ComputeMaxXSec(const Interaction * interaction) cons
           kinematics->SetKV(kKVPn, pFmax);
           // Compute the QE cross section for the current kinematics
           double xs = fXSecModel->XSec(interaction, fkps);
-          if (xs > tmp_xsec_max)
+          if (!initialized || xs > tmp_xsec_max)
           {
-             tmp_xsec_max = xs;
-             Q20 = Q2;
-             v0 =  v;
+              tmp_xsec_max = xs;
+              Q20 = Q2;
+              v0  = v;
+              initialized = true;
           }
        } // Done with v scan
     }// Done with Q2 scan
@@ -543,6 +545,7 @@ double QELEventGeneratorSM::ComputeMaxXSec(const Interaction * interaction, cons
      double xsec_max = -1;
      double tmp_xsec_max = -1;
      double Q20, v0;
+     bool initialized = false;
      const int N_Q2 = 32;
      const InitialState & init_state = interaction -> InitState();
      Target * tgt = init_state.TgtPtr();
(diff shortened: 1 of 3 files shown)


src/Physics/QuasiElastic/EventGen/QELEventGeneratorSM.cxx
```diff
@@ -486,11 +487,12 @@ double QELEventGeneratorSM::ComputeMaxXSec(const Interaction * interaction) cons
           kinematics->SetKV(kKVPn, pFmax);
           // Compute the QE cross section for the current kinematics
           double xs = fXSecModel->XSec(interaction, fkps);
-          if (xs > tmp_xsec_max)
+          if (!initialized || xs > tmp_xsec_max)
```
nusense, line 490: I'd have to think a bit more about it, but doesn't initializing `tmp_xsec_max` to `-1` guarantee that the first value of `xs` (which should never be negative, at least 0) should enter this block?

I guess there's no harm in adding this.

src/Physics/Resonance/XSection/BostedChristyEMPXSec.cxx
```diff
@@ -462,7 +459,7 @@ double BostedChristyEMPXSec::XSec(
   double nu = (Wsq - MN2 + Q2)/2./MN;
   double x  = Q2/2./MN/nu;
 
-  double sigmaT, sigmaL, F1p, R, W1;
+  double sigmaT, sigmaL, F1p, R, W1(0);
```
nusense, line 462: Good catch.  Looking at the code I'm thinking that it was left unset if `A<2 && Wsq<=1.155`.

src/Physics/Resonance/XSection/BostedChristyEMPXSec.cxx
```diff
@@ -498,21 +495,15 @@ double BostedChristyEMPXSec::XSec(
...
-      if (A<=kv.first)
-        break;
-    }
+    auto itEs = fNucRmvE.lower_bound(A);
+    if (itEs == fNucRmvE.end()) --itEs;
+    double Es = itEs->second;
+    
+    auto itKF = fKFTable.lower_bound(A);
+    if (itKF == fKFTable.end()) --itKF;
+    double kF = itKF->second;
```
nusense, line 504: There's an edge-case that might exist here (and also existed in the previous code):  if `fNucRmvE` or `fKFTable` is empty then it used to leave `Es` or `kF` undefined, but now it tries to iterate back from `end()` which is undefined behavior.  I see later in this PR you throw an exception if the map is empty but I don't know if that block of code gets called before this one.

src/Physics/QuasiElastic/EventGen/QELEventGeneratorSM.cxx
idkakorin: You're right that in the ideal case xs ≥ 0, so initializing tmp_xsec_max = -1 should ensure the first update.

However, the compiler cannot rely on that assumption and still sees a possible path where the condition is never satisfied (e.g. due to NaNs or unexpected values), leading to -Wmaybe-uninitialized warnings.

The explicit initialization/guard just makes this guarantee visible to the compiler and avoids potential UB in edge cases, without changing the intended behavior.

src/Physics/Resonance/XSection/BostedChristyEMPXSec.cxx
idkakorin: Good point. In this case the maps are guaranteed to be non-empty, as they are filled and validated in LoadConfig(), which throws if entries are missing.

These parameters are intrinsic to the Bosted–Christy model. If they are absent or corrupted, the BC fit itself becomes invalid, so LoadConfig() enforces this as part of model validation.

Therefore this code is only reached with properly initialized tables.

nusense (approved):
I'm happy with the previous responses to my feedback.   The new fix to prevent the leaking of `TLorentzVector` objects from `init_state` is probably common across many places in GENIE (because returning a pointer to manage is an odd usage pattern), so catching each is a good thing.

Issue #270, opened 2023-03-24
build_lhapdf.sh has an outdated download link


lhalic:
In build_lhapdf.sh the link for downloading LHAPDF is outdated. Anything downloaded through that link can not be unziped.

PR #482, opened 2026-04-02
update buld_lhapdf.sh to take into account that the URL on hepforge has changed and is more complex

nusense:
Also the name for v6 versions starts with LHAPDF rather than lhapdf.

This attempts to deal with outstanding issue #270


LiangLiu212:
I test your script with my environments. It works fine for 

`./build_lhapdf.sh 6.5.6`

But v5 is too old for me to compile, do we want to set v6 as default?

nusense:
@LiangLiu212 you're right, the default is ancient.  I've updated it to `6.5.6`

LiangLiu212 (approved):

PR #484, opened 2026-05-01
Bugfix: Ion pdg code for hypernuclei

LiangLiu212:
fix a typo in ion pdg code for hypernuclei

### DIFF
--- a/src/Framework/ParticleData/PDGUtils.cxx
+++ b/src/Framework/ParticleData/PDGUtils.cxx
@@ -79,7 +79,7 @@ int genie::pdg::IonPdgCode(int A, int Z, int L, int I)
 {
 // Build ion PDG code from A,Z,L,I
 
-  int ion_pdgc = 1000000000 +  L*100000000 + Z*10000 + A*10 + I;
+  int ion_pdgc = 1000000000 +  L*10000000 + Z*10000 + A*10 + I;
   return ion_pdgc;
 }
 //____________________________________________________________________________


src/Framework/ParticleData/PDGUtils.cxx
```diff
@@ -79,7 +79,7 @@ int genie::pdg::IonPdgCode(int A, int Z, int L, int I)
 {
 // Build ion PDG code from A,Z,L,I
 
-  int ion_pdgc = 1000000000 +  L*100000000 + Z*10000 + A*10 + I;
+  int ion_pdgc = 1000000000 +  L*10000000 + Z*10000 + A*10 + I;
```
nusense, line 82: The PDG code is `10𝐿𝑍𝑍𝑍𝐴𝐴𝐴𝐼` so this correction looks right.  If we could insist that GENIE is compiled with C++ 14 and beyond one could write these as `1'000'000'000` and `10'000'000` which would be much easier to read.   Do we still allow C++11?
Also, are hyper nuclei being generated?  By whom?  INCL++?  
If they're being added to the event record, are there relevant entries in `$GENIE/data/evgen/catalogues/pdg/genie_pdg_table.txt`?   If not that might be problematic.

nusense (approved):

src/Framework/ParticleData/PDGUtils.cxx
LiangLiu212: I think we can have discussion about the support of c++11. 

INCL++ will generate hypernuclei, we need to fix it before I open a PR for INCL interface. I will update the pdg table in my following PR.

PR #486, opened 2026-05-05
May 5 2026: coherent bug fix

afropapp13:
Inconsistency in the way that x is calculated and the target mass


sjgardiner:
Further investigation by @afropapp13 revealed that this is not needed after all. Closing to avoid confusion.

kjplows:
Additional plots made upon chatting with @afropapp13 : reference plots are from the Berger-Sehgal paper, [PRD 79 (2009) 053003](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.79.053003). Solid black lines are the BS xsec, dashed black lines are the Rein-Sehgal xsec, both numu + 12C (cf. Fig (3) in the paper). Red lines are the CC, NC COH xsec with this PR merged into a release-candidate branch, blue lines are GENIE v3_06_02_sbn2 tune AR25_20i_00_000.
image
image

Worth noting, the agreement between GENIE and the Berger-Sehgal calculation is not perfect: 
image

PR #489, opened 2026-05-13
when cloning a particle set LastMother to -1 in case source had two mothers

nusense:
do this at the same point where FirstMother is set to the original particle

### DIFF
--- a/src/Physics/HadronTransport/HNIntranuke2018.cxx
+++ b/src/Physics/HadronTransport/HNIntranuke2018.cxx
@@ -936,6 +936,7 @@ int HNIntranuke2018::HandleCompoundNucleus(GHepRecord* ev, GHepParticle* p, int 
                 << "*** Nothing left to interact with, escaping.";
               GHepParticle * sp = new GHepParticle(*p);
               sp->SetFirstMother(mom);
+              sp->SetLastMother(-1);
               sp->SetStatus(kIStStableFinalState);
               ev->AddParticle(*sp);
               delete sp;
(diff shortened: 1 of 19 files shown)


src/Physics/HadronTransport/HNIntranuke2018.cxx
sjgardiner, line 1: Do you understand the intended behavior of the `int mom` argument here? This seems unique to hN. All non-hN revisions look fine.

src/Physics/HadronTransport/HNIntranuke2025.cxx
sjgardiner, line 1: This is the other place that's questionable, for the same reason.

sjgardiner (commented):
Looks great @nusense, just a couple of questions for you.

nusense:
From slack discussion:
@sjgardiner All looks good, except I don't understand the hN recipe. What you have is probably OK there too, though.
@nusense 
Hmm.  No I don't really understand the original logic of it.   I think what's happening is there's split in the functionality between the cloning of the particle and the actual filling out the new particle (passed into the method).

src/Physics/HadronTransport/HNIntranuke2018.cxx
dytman: Let's be clear on this.  Setting FirstMother makes sense but LastMother less so for me.  It matters mostly for reweighting, right?  Figuring out which particles to reweight.  I've looked at a lot of event printouts and it's hard for me to see how a particle can have 2 LastMothers.  However, INujeUtils has looks for -1 as an error condition.  Maybe, this makes sense but will (almost?) never be used?

PR #491, opened 2026-05-20
Add linuxriscv64 arch to Make.include

venerd1:


nusense (approved):
This addition for linuxriscv64 seems unobjectionable.  I am curious what platform generates this ARCH

venerd1:
I am trying to compile GENIE on a Milk-V Pioneer workstation that we have at INFN-CNAF.

venerd1:
I successfully compiled GENIE on RISC-V using root v6.36 and applying #430.  Because root v6.30, the last version to support pythia6, doesn't compile on RISC-V

PR #493, opened 2026-05-22
new layout option

AndersenFred:
Included a new layout option if the environment variable `GMSGLAYOUT=FANCY` which results in a log
```
2026-05-22 10:05:11 [NOTICE] Message
```


nusense (approved):
This is a nice (fancy) improvement over the bare unix timestamps.

AndersenFred:
Didn't have a better idea for the name, but I'm open to suggestions :)

nusense:
It's not like the existing `SIMPLE` and `BASIC` are particularly descriptive.  I can never remember which is the default, or which include the timestamp.

mroda88 (approved):
looks good to me

PR #498, opened 2026-06-02
3-body phase space fixes for pion production in hA/hN 2018 and 2025

sjgardiner:
This PR updates and replaces #448 originally prepared by Gray Putnam. Changes they originally developed for the 2018 model versions have also been applied consistently to the 2025 code. The main changes involve

1. Sampling particle polar angle cosines rather than just the polar angles (a mistake present in the original code and commonly-encountered elsewhere, see https://mathworld.wolfram.com/SpherePointPicking.html).

2. Introducing a bias parameter that allows reshaping of pion production kinematic sampling away from pure 3-body phase space. The parameter is XML-configurable and defaults to a value of zero (no bias at all).

3. Updating interfaces to handle non-zero values of the bias parameter correctly.

### DIFF
--- a/config/HAIntranuke2018.xml
+++ b/config/HAIntranuke2018.xml
@@ -52,7 +52,7 @@ DelRNucleon         double  Yes   mult. factor for nucleon de-Broglie wavelength
     <param type="bool"   name="INUKE-DoCompoundNucleus"> true  </param>
     <param type="bool"   name="INUKE-XsecNNCorr">        true  </param>
 
- 
+
 
     <param type="double" name="HAINUKE-DelRPion">        0.0   </param>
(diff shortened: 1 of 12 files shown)


sjgardiner:
@dytman Sent review comments via email: "I tested this in various ways for nuclear targets and compared to old code results. No significant difference even for opening angle histograms. However, no failures so can be included."

dytman:
I tested this in various ways for nuclear targets and compared to old code results. No significant difference even for opening angle histograms. However, no failures were seen so can be included.

sjgardiner:
Re-opening after discussion with Steve, who meant to approve this rather than close it.

dytman (approved):
Tested by comparing output against existing code.  No errors but also no significant difference in observbles.  New code is more correct.

PR #499, opened 2026-06-03
hA and hN FSI bug fixes

dytman:
start of approx 3 bug fixes to hA and hN.  Since these are bug fixes, will go with both 2018 and 2025 versions.


sjgardiner:
Closing to avoid duplication. The changes have been combined with @mking99456's and merged into #503.

PR #492, opened 2026-05-21
Add configurable Q2 minimum cuts and spline consistency metadata

Pabce:
## Summary
Adds a configurable phase-space Q2 minimum cut, initially used for EM/electron interactions, with optional weak-interaction support through XML configuration or `--q2-min`.

## Main Changes
- Add `KPhaseSpaceCuts` as the central service for configurable Q2 cuts.
- Add `CommonPhaseSpaceCuts.xml` configuration XML as the single source of truth for Q2 cuts, and potential future kinematic cuts in other variables.
- Add `--q2-min` to `gevgen` and `gmkspl`.
- Apply the configured cut consistently across phase-space and affected physics modules.
- Record Q2 cut metadata in spline XML output.
- Make `gevgen` reject spline files whose recorded Q2 cut is missing or inconsistent when an effective cut is active.

## Validation
- Added `gtestKPhaseSpace` coverage for EM defaults, weak defaults, command-line overrides, probe-aware spline metadata, and XML round-tripping.

[IMPLEMENTATION_REPORT.md](https://github.com/user-attachments/files/28111713/IMPLEMENTATION_REPORT.md)

PR #502, opened 2026-06-09
Make the EM Q2 threshold (kMinQ2Limit) configurable via CommonParam

LiangLiu212:
## Motivation

The minimum Q2 for electromagnetic scattering events is a hardcoded constant in `src/Framework/Utils/KineUtils.h`:

```c++
static const double kMinQ2Limit = 0.02;  // GeV^2
```

Any study that needs a different EM low-Q2 boundary (e.g. electron-scattering validation work where the generation threshold must be varied) currently requires patching and rebuilding GENIE.

## Change

Replace the constant with a small proxy object that reads `EM-MinQ2Limit` from `CommonParam.xml` `[Lepton]` lazily on first use and caches it for the lifetime of the program:

- `KineUtils.h` — `kMinQ2Limit` becomes `extern const EMMinQ2LimitProxy`, a struct with `operator double()`, so **every existing call site compiles and behaves unchanged**.
- `KineUtils.cxx` — the proxy implementation: one read of `CommonParam[Lepton]/EM-MinQ2Limit` via `AlgConfigPool` (lazy, so the pool is guaranteed initialised), cached in a function-local static, logged once at `pNOTICE` for provenance.
- `CommonParam.xml` — adds `EM-MinQ2Limit` to the `[Lepton]` param set with the historical default `0.02`.

## Backwards compatibility

Default behaviour is identical: if the parameter (or the registry) is absent the proxy falls back to the historical `0.02 GeV^2`, and the shipped `CommonParam.xml` sets exactly that value. No call-site changes anywhere in the codebase.

## Validation

An equivalent patch has been exercised in our local electron-scattering builds: tunes overriding `EM-MinQ2Limit` between 0.02 and ~1.2 GeV^2 via per-tune `CommonParam.xml` overlays show the expected generation-level Q2 cut in both `gmkspl` splines and `gevgen` event samples (e- on C12, O(10M)-event productions), while builds without an override reproduce the previous hardcoded behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/CommonParam.xml
+++ b/config/CommonParam.xml
@@ -267,6 +267,9 @@ Or changing the name of this parameter set
 
   <param_set name="Lepton">
     <param type="bool" name="ApplyCoulombCorrection"> false </param>
+    <!-- Q^2 threshold for em scattering events, in GeV^2 (was hardcoded as
+         static const double kMinQ2Limit in src/Framework/Utils/KineUtils.h) -->
+    <param type="double" name="EM-MinQ2Limit"> 0.02 </param>
 
   </param_set>
 
--- a/src/Framework/Utils/KineUtils.cxx
+++ b/src/Framework/Utils/KineUtils.cxx
@@ -26,11 +26,36 @@
 #include "Framework/Utils/KineUtils.h"
 #include "Framework/Numerical/MathUtils.h"
 #include "Framework/Interaction/InteractionException.h"
+#include "Framework/Algorithm/AlgConfigPool.h"
+#include "Framework/Registry/Registry.h"
 
 #include <sstream>
 
 using namespace genie;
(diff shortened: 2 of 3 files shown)


LiangLiu212:
I currently work on (e, e'p) analysis. I create this PR via claude code just with prompt. 

Although I open this PR, I do not intend to include it in the v3.8 release.

jtenavidal:
This work is already ongoing in https://github.com/GENIE-MC/Generator/pull/492 and was presented last week at the GENIE core meeting. It had a lot of discussion and feedback. Pablo is already working on it.

We discussed the implementation at the GENIE meeting with specific input needs from e4nu. I would recommend avoiding duplication of efforts.

PR #506, opened 2026-07-07
Setting the SecondNucleonEmitter at run-time and enabling SRC in CC-QEL interactions

CarlosNeutrino:
## Description 
This Pull Request changes the way the SecondNucleonEmitter for SRC is configured. It adds the SecondNucleonEmitter in CC-QEL events (not configured before, as QELEventGenerator does not use FermiMover):
- Before this changes, the SecondNucleonEmitter was only read from the config/ files independently from the NuclearModel.
- The SecondNucleonEmitter depends strongly on the NuclearModel, and the code was using incompatible SecondNucleonEmitter algorithms for the specified global NuclearModel.
- StringUtils.h → used for Split
- Now, the SecondNucleonEmitter is configured at run-time depending on the global NuclearModel. In specific, the new code:
  - config/FermiMover.xml and config/QELEventGenerator.xml include two new strings ("NuclearModels" and "SecondNucleonEmitters") that read the NuclearModel along with the corresponding SecondNucleonEmitter
  - FermiMover.cxx and QELEventGenerator.cxx LoadConfig() reads the strings and sets the correct SecondNucleonEmitter.
  - The globally configured NuclearModel is read from the GlobalParameterList.

## Development
- A future improvement is to create a standalone module that runs SecondEmitter independently, so we do not duplicate the code in FermiMover and QELEventGenerator

## Notes for reviewers
- I tried to read the NuclearModel from EventGeneratorListAssembler but there was not any simpler way of doing it (neither did I manage to make it work) John Plows and I agreed to go ahead with GlobalParameterList
- I have tested that the code compiles and it runs.


config/FermiMover.xml
```diff
@@ -18,8 +18,12 @@ SecondNucleonEmitter          alg     No         SubAlg to add a second nucleon
 <alg_conf>
 
   <param_set name="Default">
-     <param type="bool"   name="KeepHitNuclOnMassShell"> false                     </param>
-     <param type="alg" name="SecondNucleonEmitter"> genie::SpectralFunction2p2h/Default </param>
+      <param type="bool"   name="KeepHitNuclOnMassShell"> false                     </param>
```
kjplows, line 21: Extra indentation... Not the end of the world but worth pointing out.

config/FermiMover.xml
```diff
@@ -18,8 +18,12 @@ SecondNucleonEmitter          alg     No         SubAlg to add a second nucleon
...
+      <param type="alg" name="NuclearModel"> genie::NuclearModelMap/Default         </param> 
+
+      <!-- Possible SecondNucleonEmitters for every different NuclearModel -->
+      <param type="string" name="NuclearModels">genie::LocalFGM;genie::FGMBodekRitchie;genie::EffectiveSF</param>
+      <param type="string" name="SecondNucleonEmitters">genie::SRCNuclearRecoil/Default;genie::SRCNuclearRecoil/Default;genie::SpectralFunction2p2h/Default</param>
```
kjplows, line 26: Just to point out some GENIE functionality here: you can use `<param type="vec-string" name="SecondNuclearEmitters" delim=";">` and access the values as a `std::vector` using `Algorithm::GetParamVect` - see `config/QELEventGeneratorSM.xml` and `src/Physics/QuasiElastic/EventGen/QELEventGeneratorSM.cxx::LoadConfig()`

src/Physics/NuclearState/FermiMover.cxx
```diff
@@ -312,10 +323,65 @@ void FermiMover::LoadConfig(void)
 
   this->GetParamDef("MomentumDependentErmv", fMomDepErmv, mom_dep_energy_removal_def);
 
-  RgKey nuclearrecoilkey = "SecondNucleonEmitter" ;
-  fSecondEmitter = dynamic_cast<const SecondNucleonEmissionI *> (this->SubAlg(nuclearrecoilkey));
-  
-  assert(fSecondEmitter);
+  // Build nuclear-model algorithm name -> SecondNucleonEmitter algorithm,
+  std::string nucl_model_names_str, fermi_mover_algids_str;
+  this->GetParam("NuclearModels",         nucl_model_names_str);
```
kjplows, line 328: See comment on the `config` about loading these directly into `std::vector<std::string>` 🙂 
(also saves you from including `StringUtils.h`)

src/Physics/NuclearState/FermiMover.cxx
```diff
@@ -312,10 +323,65 @@ void FermiMover::LoadConfig(void)
...
+    if (slash != std::string::npos) {
+      alg_name = full.substr(0, slash);
+      alg_conf = full.substr(slash + 1);  // If the slash is found, take the substring after it as the config
+    }
+
+    const Algorithm * alg = algf->GetAlgorithm(alg_name, alg_conf);
+    const SecondNucleonEmissionI * emitter =
+      dynamic_cast<const SecondNucleonEmissionI *>(alg);
+    if (!emitter) {
+    LOG("FermiMover", pWARN)
```
kjplows, line 360: I would argue making this an exception, rather than a silent continue. If this statement triggers it's either a misconfiguration or a bad runtime load, both should trigger warnings and interrupt execution. Happy to defer to @sjgardiner or @nusense on this.

src/Physics/QuasiElastic/EventGen/QELEventGenerator.cxx
```diff
@@ -404,6 +418,66 @@ void QELEventGenerator::LoadConfig(void)
     fHitNucleonBindingMode = genie::utils::StringToQELBindingMode( binding_mode );
 
     GetParamDef( "MaxXSecNucleonThrows", fMaxXSecNucleonThrows, 800 );
+
+    // Build nuclear-model algorithm name -> SecondNucleonEmitter algorithm,
+    std::string nucl_model_names_str, fermi_mover_algids_str;
+    this->GetParam("NuclearModels",         nucl_model_names_str);
+    this->GetParam("SecondNucleonEmitters", fermi_mover_algids_str);
+
+    std::vector<std::string> nucl_model_names  = genie::utils::str::Split(nucl_model_names_str, ";");
```
kjplows, line 427: Same comment as above, you can avoid this with `GetParamVect`

src/Physics/QuasiElastic/EventGen/QELEventGenerator.cxx
```diff
@@ -404,6 +418,66 @@ void QELEventGenerator::LoadConfig(void)
...
+    if(it != fSecondEmitterMap.end()) {
+        fSecondEmitter = it->second;
+        LOG("QELEvent", pINFO)
+        << "Selected " << fSecondEmitter->Id().Name()
+        << " as SecondNucleonEmitter for nuclear model " << nucl_model_name;
+    }else{
+        LOG("QELEvent", pWARN)
+        << "No SecondNucleonEmitter for nuclear model "
+        << nucl_model_name << " -- SRC recoil disabled";
+    }
```
kjplows, line 480: (This is not so much directed at this PR but applies anyway) - not a huge fan of duplicating code, but given QELEventGenerator contains duplicate code anyway, this is at least consistent. Maintenance caveat: any PRs targeting SecondNucleonEmitter have to be done in two locations in code.

kjplows (changes requested):
Overall looks good, thank you @CarlosNeutrino ! Requesting changes for the following:
1. `GetParamVect` and `vec-string` where possible
2. We should discuss making the silent `continue` statement on second-nucleon-emission load a proper Exception.

PR #511, opened 2026-07-20
fix for detecting both Geant4 v10 and v11

nusense:
Geant4 v11 splits the `libG4processes.so` library into `libG4processes_core.so` and `libG4process_hadronic.so` .  In `configure` it tests to make sure the library exists to ensure that you specified the path correctly.  Make it test for either.   Linking to an appropriate library list happens through the use of `geant4-config` which is found relative to this path.


sjgardiner (approved):
Good catch, @nusense, many thanks! We will merge this for v3.8.0.

PR #515, opened 2026-09-05
Rework RNG initialisation to only fire once so Pythia8 does not silently fail, PR to master

kjplows:
Copy of #514 , targeting master.

Summoning @nusense and @sjgardiner for review, again!

### DIFF
--- a/src/Framework/Utils/AppInit.cxx
+++ b/src/Framework/Utils/AppInit.cxx
@@ -31,7 +31,7 @@ void genie::utils::app_init::RandGen(long int seed)
 {
   // Set random number seed, if a value was set at the command-line.
   if( ! RandomGen::IsInitialized() ) {
-    if(seed > 0) { 
+    if(seed >= 0) { 
       RandomGen::Instance(seed);
     } else RandomGen::Instance();
   } else {
--- a/src/Framework/Utils/AppInit.cxx
+++ b/src/Framework/Utils/AppInit.cxx
@@ -30,9 +30,15 @@ using namespace genie;
 void genie::utils::app_init::RandGen(long int seed)
 {
   // Set random number seed, if a value was set at the command-line.
-  if(seed > 0) {
-    RandomGen::Instance()->SetSeed(seed);
-  }
+  if( ! RandomGen::IsInitialized() ) {
+    if(seed > 0) { 
+      RandomGen::Instance(seed);
+    } else RandomGen::Instance();
+  } else {
+    if(seed > 0) {
+      RandomGen::Instance()->SetSeed(seed);
+    }
+  } // if first initialisation
 }
 //___________________________________________________________________________
 void genie::utils::app_init::XSecTable (string inpfile, bool require_table)
--- a/src/Apps/gAtmoEvGen.cxx
+++ b/src/Apps/gAtmoEvGen.cxx
@@ -332,7 +332,7 @@ int main(int argc, char** argv)
   // Parse command line arguments
   GetCommandLineArgs(argc,argv);
 
-  // Seed should always be initialised first.
+  // Seed should always be initialised first to ensure all RNG seeds are consistent.
   utils::app_init::RandGen(gOptRanSeed);
 
   if ( ! RunOpt::Instance()->Tune() ) {
@@ -341,7 +341,7 @@ int main(int argc, char** argv)
   }
   RunOpt::Instance()->BuildTune();
 
-  // Iinitialization of random number generators, cross-section table, messenger, cache etc...
+  // Initialization of cross-section table, messenger, cache etc...
   utils::app_init::MesgThresholds(RunOpt::Instance()->MesgThresholdFiles());
   utils::app_init::CacheFile(RunOpt::Instance()->CacheFile());
   utils::app_init::XSecTable(gOptInpXSecFile, true);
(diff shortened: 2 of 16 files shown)


sjgardiner:
@nusense Please have a look when you get back to work. I've already merged this into a release-candidate branch that John is actively exercising. We made a separate PR to `master` to keep around for review.

nusense (commented):
I'm still working through all the typical scenarios, of which I see five:

<pre>
(1)   utils::app_init::RandGen(<value>) called <value> non-zero
      [tune is configured]
      rndm = RandomGen::Instance()->RndFlux().Rndm()
(2)   utils:app_init::RandGen(0)
      [tune is configured]
      rndm = RandomGen::Instance()->RndFlux().Rndm()
(3)   [tune is configured]
      utils::app_init::RandGen(<value>) called <value> non-zero
      rndm = RandomGen::Instance()->RndFlux().Rndm()
(4)   [tune is configured]
      utils:app_init::RandGen(0)
      rndm = RandomGen::Instance()->RndFlux().Rndm()
(5)   utils::app_init::RandGen() never called; [tune is configured]
      rndm = RandomGen::Instance()->RndFlux().Rndm()
</pre>

I'm still working them through in my head.  I'm also working out the use case for `static RandomGen * Instance(long int seed)`.   I'm not sure where `Instance(seed)` gets constructed rather than `Instance()`.

Would it be possible to incorporate the changes in #516 into this PR, so it would be easier to see the totality of the changes.

I think I copied:
<pre>
  // sync GENIE and PYTHIA8 seeds
  RandomGen * rnd = RandomGen::Instance();
  long int seed = rnd->GetSeed();
  fPythia->readString("Random:setSeed = on");
  fPythia->settings.mode("Random:seed", seed);
</pre>
from Costas' first attempt at supporting Pythia8, in `src/Physics/Hadronization/Pythia8Hadro2019.cxx` circa `R-3_04_00`.
Which is why the random seed is initialized in the algorithm.   I'm not sure why it was done this way, but I blindly followed it.

But it would make sense to also have it in `RandomGen::SetSeed()` (around lines 22,  115) similarly to how the pythia6 seed is set .
<pre>
#ifdef __GENIE_PYTHIA8_ENABLED__
#include "Framework/Utils/Pythia8Singleton.h"
#endif
...
#ifdef __GENIE_PYTHIA8_ENABLED__
    Pythia8::Pythia* gPythia = Pythia8Singleton::Instance()->Pythia8();
    gPythia->readString("Random:setSeed = on");
    gPythia->settings.mode("Random:seed", seed);
    gPythia->init();
#endif
</pre>
One probably needs `init()` there to ensure that it was picked up in case where the tune was constructed prior to the `SetSeed()` scenarios 3-5.  Even if we fix the order in GENIE supplied apps, that doesn't guarantee that user generated alternative applications won't have it the same as they used to be.   With this addition, it _might_ make sense to remove the `Random:[set]Seed` lines from the 3 algorithms.  Those algorithms also set other things, so I think those must stay with their accompanying `init()`

nusense:
Hmm.   I may have remembered why the direct Pythia8 call wasn't in `RandomGen`.   I think this introduces a circular dependency, where `Framework/Numerical/RandomGen` now depends on `Framework/Utils/Pythia8Singleton`, but various classes in `Framework/Utils` rely on a number of classes in `Framework/Numerical`.  [sigh].

nusense:
Is `RandomGen` a "Numerical" thing, or a "Util" thing :-)   We could consider moving it -- with lots of changes to header include lines *everywhere*.  That's probably too big a "lift" as it might also affect user code.

kjplows:
Yep moving RandomGen seems quite a heavy lift indeed! 

So, I think that regardless of coverage (seed instantiation vs tune configuration), having _some_ synchronisation of the seed is better than having none, and that seems to be resolved without adding the calls to the `RandomGen` code. So, perhaps a middle ground is to keep the c'tor with seed, and not explicitly call init() to Pythia8 in RandomGen? 🤔

nusense:
I think without the call to `init()` in `RandomGen` there's not guarantee that `init()` will be called appropriately  if the tune is created before the `utils::app_init::RandGen(seed)` is called in the app.  And then it sits in the settings database but isn't actually seeded into the Pythia RNG.

kjplows:
Just addressing a comment from @nusense solving the innocuous warning Pythia8 fires on early initialisation in RandomGen - Pythia is not the primary event generator so warning that no processes are configured is benign.

src/Framework/Utils/AppInit.cxx
```diff
@@ -30,9 +30,15 @@ using namespace genie;
...
   // Set random number seed, if a value was set at the command-line.
-  if(seed > 0) {
-    RandomGen::Instance()->SetSeed(seed);
-  }
+  if( ! RandomGen::IsInitialized() ) {
+    if(seed >= 0) { 
+      RandomGen::Instance(seed);
+    } else RandomGen::Instance();
+  } else {
+    if(seed > 0) {
```
nusense: Does this need to be `seed >= 0` here as well?

nusense (approved):
I'm approving w/ one comment that probably should be changed.

src/Framework/Utils/AppInit.cxx
kjplows: Good catch! Thanks, updating now.

PR #519, opened 2026-09-17
Separate out hadron fates for Intranuke, Intranuke 2018, and Intranuke 2025

kjplows:
The enums and classes for the hadron fates differ between Intranuke (has elastic enabled for hA), and Intranuke 2018/2025 (identical copies of the same class). All are scoped in the global `genie` namespace. All had the same C++ preprocessor include guard. 
The three classes now have separate include guards and enums. Code invoking the various implementations of Intranuke is now required to use its own Intranuke's fates, and includes.


dytman:
Why was this done?  Yes, Robert doesn't like what was done in the past.  Original version had extra fate which is no longer needed.  That version will disappear soon.  2018 and 2025 are designed to have same set of fates but they will be populated in different ways.

kjplows:
Hi @dytman ! So as part of validation tests for the 3.8 release candidate, I'm trying to make ReWeight work with hA2025 loaded (see [ReWeight/#54](https://github.com/GENIE-MC/Reweight/pull/54)). Part of the issue I found was conflicting header guards in the hA2018 and hA2025 fates, which is why I reworked part of Generator to explicitly separate them. It's still a bit finicky to get hA2025 to play well there..

For now, I'm marking the PR as a draft again, while development is active. Would appreciate your insights on this!

PR #520, opened 2026-09-18
Tweak CommonParam.xml GVLD-Emax from 1000.000 to 1000.010

nusense:
Traditionally community spline files go up to 1000 GeV.  I've previously seen some cross-section evaluations set to 0 at 1000 GeV.  This messes up the TSpline3 evaluation for the last three energy points because it has to take a sharp turn over to hit the zero at that last knot.  
  
 I surmise that these zero xsec values where due to the algorithm doing a `epoint >= emax` test, but I could never find it in the code.  As an alternative we'll change all instances of `GVLD-Emax` set to `1000.000` in any `CommonParam.xml `file to `1000.010` to hopefully avoid this issue.


nusense:
mixed two sets of changes

